PR #195, opened 2026-06-27
fix: improve manga import review

### LINKED ISSUE #186: Import manga improvement/question
**Import question**
What happens if comicarr can't infer the chapter number of a file, as currently matching is done on a group base instead of per file basis?

**Describe the request**
The request is currently is 2 fold as there is a config option that changes this. See [IMP_SERIESFOLDERS](https://comicarr.com/docs/guides/importing#import-settings) from the docs.
1. If the series folder is used: Use the folder as the manga name, also use the match function to link to this. And then list the files in the entry, with the chapter number inferred from the file name. But also allow to manually specify the chapter number.
2. If the series folder is not used: Have one entry per file where. Where matching the show for the single file and also display the inferred chapter number for the file and allow editing in case it can't be inferred automatically.

This could also be done automatically of course depending of found file is in a sub-directory of the import directory or not, and then the whole import option in the config would not be necessary.

**How did you come up with the idea?**
This would match the import operation more to how sonarr does manually import yet adapted for manga style import.

**Additional information**
Currently files get grouped together in the wrong way and are all grouped in to one entry to be matched.
In the import directory I had the following folder/file structure:
```
- Manga A/
-- chapter 1.cbz
-- chapter 2.cbz
- Manga B/
-- chapter 1.cbz
-- chapter 2.cbz
```
But in the UI they are grouped together in to one entry to match. And not 2 separate do the matching on.

frankieramirez:
## Summary

Manga imports now preserve the user's actual folder and file intent instead of collapsing chapter-only filenames into ambiguous review rows. Nested folders become folder review groups, root files remain separate single-file groups, inferred chapters are visible and editable per file, and matching uses the reviewed per-row chapter values during finalization.

The Import page now opens on the review queue when pending work exists. It shows accurate group/file counts, clearer folder vs single-file labels, inline chapter/issue save states, safer match handoff context, and keeps library scans plus the watched inbox below as setup and maintenance tools.

Fixes frankieramirez/comicarr#186.

## Design Notes

- `GET /api/import` keeps `pagination.total` as the group count and adds `summary.group_count` plus `summary.file_count` for accurate page copy.
- Import Inbox grouping uses stable `DynamicName` keys so folder groups and root single files stay distinct through pending review.
- Chapter edits update only `importresults.IssueNumber`; no new columns or schema migration are required.
- Match finalization resolves issue/chapter IDs per import row, so mixed chapter groups do not reuse one shared issue ID.
- The review UI treats scan/setup surfaces as secondary when pending imports exist, while preserving the existing scan and inbox flows.

## Validation

- `.venv/bin/python -m pytest tests/unit/test_importinbox.py tests/unit/test_manga_parser.py tests/unit/test_series_import_matching.py -q` — 80 passed
- `cd frontend && npm run lint`
- `cd frontend && npm run typecheck`
- `cd frontend && npm run test:run` — 62 passed
- `cd frontend && npm run build`
- `git diff --check`
- Captured local production-build screenshots for desktop collapsed, expanded, saving, saved, save-error, empty, mobile expanded, mobile empty, and mobile action-scroll states.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

## Summary by CodeRabbit

* **New Features**
  * Added inline editing for issue/chapter numbers on pending import files.
  * Added clearer import group summaries, including separate counts for groups and files.
  * Added a “Scan inbox now” action when no pending imports are found.

* **Bug Fixes**
  * Improved import grouping and selection handling for folders and single files.
  * Better chapter detection for filenames that only include chapter information.
  * Updated match and review labels to show more helpful context and confidence messages.

* **UI Improvements**
  * Added tooltips and clearer buttons across import review actions.

### DIFF
--- a/comicarr/app/series/queries.py
+++ b/comicarr/app/series/queries.py
@@ -13,7 +13,7 @@ Series domain queries — comics, issues, annuals, importresults tables.
 Uses SQLAlchemy Core via the existing db module.
 """
 
-from sqlalchemy import delete, func, select
+from sqlalchemy import delete, func, or_, select
 
 from comicarr import db
 from comicarr.app.core.database import paginated_query  # noqa: F401 — re-exported
@@ -249,6 +249,11 @@ def get_wanted_annuals():
 def get_import_pending(limit=50, offset=0, include_ignored=False):
     """Get pending import files grouped by DynamicName/Volume with pagination."""
     ir = t_importresults
+    group_key_expr = func.coalesce(
+        func.nullif(ir.c.DynamicName, ""),
+        func.nullif(ir.c.ComicName, ""),
+        ir.c.impID,
+    )
 
     base_conds = [
         (ir.c.WatchMatch.is_(None)) | (ir.c.WatchMatch.like("C%")),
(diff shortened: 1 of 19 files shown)


frankieramirez:
Addressed the CodeRabbit feedback in `138f0347`:

- `get_issue_id_for_import` now resolves `ChapterNumber` first and only falls back to `Issue_Number` if no chapter match exists.
- Folder import grouping now uses a stable path-derived `folder:` key while keeping the readable folder name as the review display name, with collision coverage for similar folder names.
- Expanded import file rows now track the last successfully saved chapter/issue value locally, avoid duplicate saves before refetch, and treat Escape as a true cancel.
- Bulk selection is re-derived and emitted when the `imports` payload refreshes, so parent bulk actions do not keep stale file IDs.
- Folder-context numeric chapter files like `001.cbz` / `002.cbz` now use `Chapter`/`Chapters` labels and have regression coverage.
- Import page review summaries now show error-specific copy instead of fallback `0 groups · 0 files` counts when `/api/import` fails.

Validation run locally:
- `.venv/bin/pytest tests/unit/test_importinbox.py tests/unit/test_series_import_matching.py`
- `.venv/bin/ruff check comicarr/`
- `.venv/bin/ruff format --check comicarr/`
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run test:run`
- `npm run build`

PR #539, opened 2026-08-04
feat: Bound the needs-attention band and add a triage route

frankieramirez:
Activity: bound the needs-attention band and give triage its own route

The band had no cap — one row per unresolved journal row — so a restart-replay burst pushed the timeline off the page. This implements the Variant C decision from #526, which was prototyped but never built.

- **Grouping (#524):** the server groups the band by `(comicid, base_reason)` from `payload_json`, singleton on `release_key` when there's no comicid. No library join, never keyed on `comicname`. `list_attention_groups` and `count_attention_groups` share one builder, so the band and the status count can't drift; `attention` is now a group count with `attention_members` beside it. Operator phrases live in `activity/reasons.py` next to the base-token matching that picks them, so wording and matching can't drift either.
- **Band + route (#526):** the band becomes a bounded preview that routes — 5 cards in one fixed-height row, `×N` volume, stage colour, folding into the new `/activity/attention` via "See all N →". That's a route, not a fourth tab, so the documented tab order is untouched. Triage owns the workspace: stage/age facets, search, selection, and cap surfacing. The `?variant=` prototype is deleted, as its own resolution required.
- **Bulk + rename (#525):** `POST /api/downloads/needs-attention/batch` fans one action over many keys — best-effort per row, cap 25 newest-first, multi-status body reported as a summary toast. `ignore` is renamed `stop_wanting` end-to-end; the durable `ignored` stamp keeps its spelling, since renaming a persisted status would re-admit every stamped row. Two or more issues now require a consequence confirmation.

## Verification

Run against a throwaway instance seeded with 300 rows in the production shape from #521 (Looney Tunes ×173 and the long tail): 300 rows collapse to 20 cards, the band holds one card row, and the timeline sits directly beneath it. A live batch action dropped the counts to 19 · 299 consistently across the page header, the band, and the status bar.

2070 backend tests and 324 frontend tests pass; the full `npm run lint` lane is green. `docs/architecture/activity-center.md` is updated for the grouping contract, the triage route, and the action rename.

## Not included

**#523's admission rewrite.** Its 8 excluded reason tokens each owe a reconciliation (blocklist + re-want), and shipping the exclusion without it would strand ~195 issues silently at `Snatched`. It also wants its own ADR. Grouping alone takes a 389-row band to ~20 groups; #523 would first drop it to ~204 rows. `reasons.py` is where those verdicts belong when it lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QTzNNDvchxLBATtCsZXGQE

## Summary by CodeRabbit

* **New Features**
  * Added a grouped Needs Attention band above the Activity timeline, showing up to five newest issue groups with severity, reasons, and issue totals.
  * Added a dedicated Needs Attention page with filtering, searching, selection, expansion, and per-group actions.
  * Added bulk actions for up to 25 issues with success, partial-result, and error feedback.
  * The status indicator now links directly to Needs Attention.
* **Improvements**
  * Renamed “Ignore” to “Stop wanting,” with confirmation for multiple issues.
  * Activity wording is now clearer and consistent across views.


frontend/src/components/activity/timeline/types.ts
```diff
@@ -62,21 +62,47 @@ export interface TimelineEvent {
...
+  base_reason: string | null;
+  reason_phrase: string;
+  member_count: number;
+  newest_updated_at: string;
+  oldest_updated_at: string;
+  stage: BandStage | string;
+  /**
+   * Stage intersection across members — empty for mixed-stage groups, whose
+   * rows are resolved by selecting members instead.
+   */
```
frankieramirez, line 103: Addressed in b51dccc1 — and my earlier decline was wrong.

I declined this on the grounds that mixed groups cannot occur: reason → stage is a function, so every member of a `(comicid, base_reason)` group shares a stage. That is true of the **writers**, and I verified it by AST-walking every `mark_failed`/`mark_manual_review` call site. But it is the wrong question. Grouping runs over rows that already exist, and `_purge_pipeline_journal` prunes only `post_processed` rows and **resolved** failed/manual_review terminals after 365 days — unresolved band rows are never pruned. So a journal written by an older Comicarr can hold a group whose rows sit at different stages today. The state is reachable in real data; I reasoned about the code and not the database.

What changed:

- `AttentionMember` now carries `available_actions`, derived server-side from its own stage (`grouping.member_actions`).
- Triage selects by `release_key`, not `group_key`. A group checkbox is shorthand for all of its members and goes indeterminate when only some are picked; each member row has its own checkbox.
- `sharedActions` intersects **selected rows** rather than selected groups, so a cross-stage selection still offers whatever both stages admit — and says so plainly when nothing does, instead of rendering an empty button row.
- Mixed groups open expanded and label each row's stage, because picking rows is the only way to act on them; hiding them behind `show N issues` is what made the group read as a dead end.

Verified live against a seeded legacy-shaped mixed group (1 `failed` + 2 `manual_review` under one comicid/reason): selecting only the failed row offers **Retry**; selecting all three collapses to **Stop wanting**, the single action both stages admit. Tests cover both paths plus the group-level DTO.

`test_reason_to_stage_is_a_function` stays but is reframed — it guards the UX property that newly written groups are single-stage, not a correctness invariant the UI may lean on.

PR #807, opened 2026-09-01
fix(search): try each candidate as itself, not as the first result

### LINKED ISSUE #808: One previously-failed release rejects every alternative candidate for the same issue
## Summary

`verification()` iterates over the candidates an issue matched, but hands `searcher()` the **whole** candidate list. `searcher()` reads every field it needs off `comicinfo[0]`, so each candidate is described by candidate 0.

The visible effect is that **one previously-failed release rejects every alternative for that issue**. `failed_check()` is called with candidate 0's `nzbid` on every pass, so once a release is in the `failed` table the search rejects candidate 0, then rejects candidates 1..n *as candidate 0*, and ends with `Could not find Issue N` — while a well-seeded alternative sits in the same result set.

## Root cause

```python
# comicarr/search.py, verification()
for verified in verified_matches:
    ...
    searchresult = searcher(
        verified["nzbprov"],
        nzbname,
        verified_matches,          # <- whole list, not the current candidate
        links,
        verified["IssueID"],
        ...
    )
```

`searcher()` unpacks its `comicinfo` argument positionally:

```python
ComicName   = comicinfo[0]["ComicName"]
IssueNumber = comicinfo[0]["IssueNumber"]
nzbid       = comicinfo[0]["nzbid"]
```

So `nzbid`, `size`, `nzbtitle` and `pack_numbers` all come from the first match regardless of which candidate the loop is on. `nzbid` is the one with teeth, because it is what `FAILED_DOWNLOAD_CHECKER` matches on.

Everything *after* the loop already uses the right candidate via `verified_matches[verified_index]` (`["pack"]`, `["nzbid"]`, `["entry"]`, `["pack_numbers"]`, …). Only the `searcher()` call ignores it.

## Observed

Four distinct candidates matched one issue, and all four were checked under the first candidate's guid:

```
_match_entry : title_id: ...guid=c2569165793b89a4a1f396377b266147
_match_entry : title_id: ...guid=66a98248e1252d915890382acd51fb99
_match_entry : title_id: ...guid=43cd0587987b2d4cde2f996724b5d157
_match_entry : title_id: ...guid=9925a0f01f4448a1b550d3c07217cd58
...
failed_check : prov : NZBGeek (newznab)[c2569165793b89a4a1f396377b266147]   x4
search_init  : Could not find Issue 85 of Invincible (2003) using NZBGeek [api]
```

Candidate 2 (`66a98248…`) was a healthy release from a different scan group — the same group 85 of the 136 issues already in that library came from. It was never tried.

## Reproduce

1. Have a release for some issue recorded in the `failed` table (any real failed download; see #805 for one way this ends up empty on an install)
2. Ensure the provider offers more than one candidate for that issue
3. Search the issue

Expected: the failed release is skipped and the next candidate is sent.
Actual: every candidate is checked as the failed one, and the search ends `Could not find Issue N`.

## Why this survived

It only bites once the `failed` table is non-empty. On an install where `markFailed()` never fires, `failed_check()` always returns `Good`, the wrong `nzbid` is never consulted destructively, and the bug is invisible. On my install that was the case for the app's entire history — the `failed` table had 0 rows because of #805, so fixing that is what exposed this.

## Proposed fix

Pass the candidate currently being tried:

```python
-                    verified_matches,
+                    [verified],
```

## Verified

Applied to `v0.38.2` on top of #806 and deployed against the reproducing install. The id now advances with the candidate and the fallback is taken:

```
searcher : nzbid: c2569165793b89a4a1f396377b266147
           [FAILED_DOWNLOAD_CHECKER] result has a status of FAIL ... continuing search
searcher : nzbid: 66a98248e1252d915890382acd51fb99
           [FAILED_DOWNLOAD_CHECKER] Successfully marked this download as Good
           This is not in the failed downloads list. Will continue with the download.
```

That issue downloaded and imported. Across the series, seven issues that had been failing repeatedly went skip → advance → alternative → imported, and the library went from 136/145 to 144/145. The one remaining gap is a genuinely misnamed release, unrelated to this.

I cross-checked the selections against a direct newznab query: the release Comicarr picked after the fix is the same one the indexer ranks with ~1,400 grabs, versus ~20 for the failed one it had been re-selecting.

PR follows with a regression test that fails on `main` with `['empire-id', 'empire-id']` — the wrong id repeated.

azuretek:
Fixes #808.

### Problem

`verification()` loops over `verified_matches`, but passes the **whole list** into `searcher()`. `searcher()` reads every field it needs off `comicinfo[0]`:

```python
ComicName   = comicinfo[0]["ComicName"]
IssueNumber = comicinfo[0]["IssueNumber"]
nzbid       = comicinfo[0]["nzbid"]
...
```

So every candidate is described by candidate 0 — including `nzbid`, `size`, `nzbtitle` and `pack_numbers`.

The visible effect is in failed-download handling. `failed_check()` is called with candidate 0's `nzbid` for **every** candidate, so once a release is recorded in the `failed` table it rejects not only itself but every alternative the provider offered for that issue. The search then ends in `Could not find Issue N` while a perfectly good release sits in the same result set.

### Observed

Four distinct candidates matched for one issue, and all four were checked under the first candidate's guid:

```
_match_entry : title_id: ...guid=c2569165793b89a4a1f396377b266147
_match_entry : title_id: ...guid=66a98248e1252d915890382acd51fb99
_match_entry : title_id: ...guid=43cd0587987b2d4cde2f996724b5d157
_match_entry : title_id: ...guid=9925a0f01f4448a1b550d3c07217cd58
...
failed_check : prov : NZBGeek (newznab)[c2569165793b89a4a1f396377b266147]   x4
search_init  : Could not find Issue 85 of Invincible (2003) using NZBGeek [api]
```

The indexer had a well-seeded alternative release for that issue (candidate 2). It was never tried.

After the change, the id advances with the candidate and the fallback is taken:

```
searcher : nzbid: c2569165793b89a4a1f396377b266147
           [FAILED_DOWNLOAD_CHECKER] result has a status of FAIL ... continuing search
searcher : nzbid: 66a98248e1252d915890382acd51fb99
           [FAILED_DOWNLOAD_CHECKER] Successfully marked this download as Good
           This is not in the failed downloads list. Will continue with the download.
```

The issue downloaded and imported.

### Note on why this survived

It only bites once the `failed` table is non-empty. On an install where `markFailed()` never fires, `failed_check()` always returns `Good`, so the wrong id is harmless and invisible.

### Change

`verified_index` already tracks the current candidate for the code after the loop (`verified_matches[verified_index]["pack"]`, `["nzbid"]`, `["entry"]`, …). Only the `searcher()` call ignored it. Pass the candidate itself.

### Tests

`tests/unit/test_search_verification_candidate.py` — two cases, both of which fail on `main`:

- the `nzbid` seen by `searcher()` must advance with the candidate (on `main`: `['empire-id', 'empire-id']`)
- a failed first candidate must not prevent the second from being sent

Full unit suite passes (2816), `ruff` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release verification so each candidate is checked individually.
  * Correctly evaluates alternative candidates when the first release fails verification.

* **Tests**
  * Added coverage confirming candidate-specific checks and successful fallback to valid alternatives.

### DIFF
--- a/comicarr/search.py
+++ b/comicarr/search.py
@@ -1514,21 +1514,29 @@ def verification(verified_matches, is_info):
     done = False
     verified_index = 0
     if verified_matches != "no results":
-        for verified in verified_matches:
+        # verified_index has to track the candidate actually in hand: the post-loop
+        # block reads verified_matches[verified_index] for the pack check, the nzbid
+        # it hands nzblog(), and the snatch notification. search_filer also emits
+        # alt_match entries with downloadit=False into this same list, and those were
+        # skipped without advancing the index -- so any candidate sent after one of
+        # them was logged under an earlier entry's nzbid.
+        for verified_index, verified in enumerate(verified_matches):
             if verified["downloadit"]:
                 try:
                     if verified["chkit"]:
                         helpers.checkthe_id(ComicID, verified["chkit"])
                 except Exception:
                     pass
-                nzbname = nzbname_create(is_info["nzbprov"], info=verified_matches, title=verified["ComicTitle"])
+                # nzbname_create() reads ComicName, IssueNumber and comyear off
+                # info[0] for the blackhole name, so it needs the candidate being
+                # tried for the same reason searcher() below does.
+                nzbname = nzbname_create(is_info["nzbprov"], info=[verified], title=verified["ComicTitle"])
                 if nzbname is None:
                     logger.error(
                         "[NZBPROVIDER = NONE] Encountered an error using given "
                         "provider with requested information: %s. You have a blank "
                         "entry most likely in your newznabs, fix it & restart Comicarr" % verified
                     )
-                    verified_index += 1
                     continue
                 try:
                     links = {"id": verified["entry"]["id"], "link": verified["entry"]["link"]}
@@ -1561,7 +1569,6 @@ def verification(verified_matches, is_info):
                     ]
                 ):
                     is_info["foundc"]["status"] = False
-                    verified_index += 1
                     continue
                 elif any(
                     [
@@ -1573,7 +1580,6 @@ def verification(verified_matches, is_info):
                     ]
                 ):
                     is_info["foundc"]["status"] = False
-                    verified_index += 1
                     return is_info
 
                 searchresult["nzbid"]
--- a/comicarr/search.py
+++ b/comicarr/search.py
@@ -1537,7 +1537,13 @@ def verification(verified_matches, is_info):
                 searchresult = searcher(
                     verified["nzbprov"],
                     nzbname,
-                    verified_matches,
+                    # searcher() reads every field off comicinfo[0], so it has to be
+                    # handed the candidate currently being tried. Passing the whole
+                    # verified_matches list pins it to the first candidate, so the
+                    # failed-release check (and the nzbid, size and title it logs)
+                    # describe candidate 0 no matter which one is in hand -- meaning
+                    # one previously-failed release rejects every alternative.
+                    [verified],
                     links,
                     verified["IssueID"],
                     verified["ComicID"],
--- a/tests/unit/test_search_verification_candidate.py
+++ b/tests/unit/test_search_verification_candidate.py
@@ -1,3 +1,12 @@
+#  Copyright (C) 2025–2026 Comicarr contributors
+#
+#  This file is part of Comicarr.
+#
+#  Comicarr is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+
 """verification() must hand searcher() the candidate it is currently trying.
 
 searcher() reads ComicName, nzbid, size, nzbtitle and friends off comicinfo[0].
@@ -10,14 +19,14 @@ alternative the provider offered.
 import comicarr.search as search
 
 
-def _candidate(nzbid, title):
+def _candidate(nzbid, title, downloadit=True, comyear="2011"):
     return {
-        "downloadit": True,
+        "downloadit": downloadit,
         "chkit": None,
         "ComicTitle": title,
         "nzbid": nzbid,
         "nzbtitle": title,
-        "comyear": "2011",
+        "comyear": comyear,
         "nzbprov": "NZBGeek",
         "tmpprov": "NZBGeek (newznab)",
         "IssueID": "1001",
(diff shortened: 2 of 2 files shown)


comicarr/search.py
```diff
@@ -1537,7 +1537,13 @@ def verification(verified_matches, is_info):
...
                     verified["nzbprov"],
                     nzbname,
-                    verified_matches,
+                    # searcher() reads every field off comicinfo[0], so it has to be
+                    # handed the candidate currently being tried. Passing the whole
+                    # verified_matches list pins it to the first candidate, so the
+                    # failed-release check (and the nzbid, size and title it logs)
+                    # describe candidate 0 no matter which one is in hand -- meaning
+                    # one previously-failed release rejects every alternative.
+                    [verified],
```
frankieramirez, line 1554: Now that fallback candidates can actually succeed, the post-loop code at the bottom of `verification()` can name the wrong release. `verified_index` only advances inside the `if verified["downloadit"]:` branch (line 1518), and `search_filer` emits alt_match entries with `downloadit=False` in the same list. Put one of those ahead of the candidate that gets sent and `verified_matches[verified_index]` in the pack branch and the `nzblog(id=...)` call points at the earlier entry.

Easiest fix I can see is `for verified_index, verified in enumerate(verified_matches):` and dropping the manual `+= 1` lines, plus a test with a leading `downloadit=False` candidate asserting nzblog gets the accepted nzbid.

frankieramirez, line 1554: nit: `nzbname_create(..., info=verified_matches, ...)` a few lines up (1524) has the same shape. Its blackhole branch reads `info[0]`, so under blackhole a later candidate's nzbname still comes from candidate 0. `info=[verified]` would match this line. Only comyear really differs between candidates, so not blocking.

frankieramirez (commented):
Fix checks out, and I confirmed the two tests fail on main. Two follow-ups in the same loop, both on lines the diff doesn't touch so I'm leaving them here.

comicarr/search.py
azuretek: No problem — good catch, and you were right that it's a real one, not a theoretical one. `search_filer` really does put `downloadit=False` alt_match entries in the same list, so this fires the moment a fallback is reachable, which is exactly what the rest of this PR unlocks.

Fixed in c03356f, using your suggestion: `for verified_index, verified in enumerate(verified_matches):` with the three manual `+= 1` lines dropped. The `verified_index = 0` initialiser stays so an empty list still can't `NameError` on the post-loop block.

Test added as you described — `test_skipped_candidate_does_not_misname_the_snatch`, with a leading `downloadit=False` candidate ahead of the accepted one. Without the fix it fails with:

```
AssertionError: nzblog was given the skipped candidate's nzbid: 'alt-id'
assert 'alt-id' == 'minutemen-id'
```

It also asserts the year handed to `notify_snatch`, since that reads `verified_matches[verified_index]["comyear"]` off the same index.

Worth noting the blast radius was wider than just a mislabelled log line: `nzblog` is what post-processing matches a completed download back to, so the accepted release was being recorded under a different entry's nzbid.

Full suite green (2816 passed), `ruff check` / `ruff format --check` / `lint:guards` clean.

azuretek: Easy change — done in the same commit, check it out now!

You're right that it's the same shape: `nzbname_create` reads `ComicName`, `IssueNumber` and `comyear` off `info[0]` for the blackhole name, so it had the identical candidate-0 pin. It's `info=[verified]` now, matching the `searcher()` call below it.

Agreed it was only `comyear` that actually differed between candidates in practice, but leaving one of the two calls reading the whole list is the kind of thing that reads as intentional later, so it seemed better to make them consistent while we're in here.

tests/unit/test_search_verification_candidate.py
```diff
@@ -0,0 +1,93 @@
+"""verification() must hand searcher() the candidate it is currently trying.
```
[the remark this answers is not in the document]
azuretek: No problem — added, GPL header now sits above the module docstring in c03356f, matching the other test modules.

azuretek:
Closing this — it's already in `main`.

#806 carried this commit (that's the overlap I flagged on that PR), so merging it took both changes together. `48aab84` landed the search fix and all three of its tests; `comicarr/search.py` and `tests/unit/test_search_verification_candidate.py` on `main` are now byte-identical to this branch, and `verified_index += 1` is gone from the file entirely.

Worth being explicit that reopening or merging this would now be **harmful** rather than redundant: the branch predates #828 and the 0.38.3 release, so its diff against current `main` is −439 lines of reversion — it would undo the metatag `.cbr` fix and roll the version back.

Thanks for the `verified_index` catch — that one was a real bug hiding behind a fix, and it went in with the rest. Your `nzbname_create(info=...)` nit is in there too.

The changeset on #806 covers both halves, so nothing is missing from the changelog: it's in `0.38.3` as "A previously failed release no longer rejects every alternative for the same issue."

PR #813, opened 2026-09-01
fix(postprocess): resolve NZBGet downloads under the category subdirectory

### LINKED ISSUE #809: Post-processing reports "Download directory not found" for a completed NZBGet download
## Summary

When NZBGet is configured with a category, it writes completed downloads to
`<MainDir>/<Category>/<NzbName>`. Comicarr's post-process input resolution joins only
`<nzbget_directory>/<nzb_name>`, so it looks one level too high, cannot find the
download, and aborts.

The download itself is complete and healthy on disk.

## Environment

- Based on `main` @ `01fbeb5`
- NZBGet, with a category configured (`nzbget_category = Comics`)

## Observed

```
[POST-PROCESSING] Processing manga download for: One-Punch Man
[POST-PROCESSING] Download directory not found:
    /azuredata1/usenet/completed/One-Punch.Man.v01.2014.Digital.F.BlackManta-Empire
```

The file was actually at:

```
/azuredata1/usenet/completed/Comics/One-Punch.Man.v01.2014.Digital.F.BlackManta-Empire
```

## Root cause

The NZBGet branch of input resolution builds the folder as a plain join:

```python
folder = os.path.join(context.nzbget_directory, context.nzb_name)
```

`nzbget_directory` is NZBGet's `MainDir`/completed path. With a category set, NZBGet
files the download under a subdirectory named for the category, so the joined path does
not exist.

Notably, the **SABnzbd branch in the same function already handles this shape** — it
builds candidate paths and falls back rather than assuming a single layout. The NZBGet
branch simply never got the same treatment.

## Impact

Not manga-specific. Any Comicarr install where NZBGet has a category configured hits
this on the NZBGet post-processing hook path.

Where Folder Monitor is also enabled and watching the category directory, the file is
still eventually imported, which can mask the failure — but the hook aborts, and on our
install the aborted run left an obligation stranded mid-flight (see the companion
recovery issue).

## Proposed fix

Resolve the NZBGet download by trying the category subdirectory as well as the plain
path, mirroring the candidate-and-fallback pattern the SABnzbd branch in the same
function already uses. Prefer an existing plain path when present so that installs
without a category are unaffected.

azuretek:
Fixes #809.

## Problem

NZBGet writes completed downloads to `<MainDir>/<Category>/<NzbName>` when a category is
configured. Input resolution joined only `<nzbget_directory>/<nzb_name>`, so it looked
one level too high and aborted with `Download directory not found` on a download that was
complete and healthy on disk.

## Change

Resolve the NZBGet download against candidate paths — the plain path and the category
subdirectory — instead of assuming a single layout. This mirrors the candidate-and-
fallback pattern the **SABnzbd branch in the same function already uses**, so it is a
consistency fix rather than a new approach.

An existing plain path still wins, so installs without a category configured are
unaffected.

## Test

Three regression tests in `tests/unit/test_postprocess_input_resolution.py`:

- resolution finds the download under the category subdirectory
- an existing plain path is still preferred (the fallback does not steal a download that
  is already where we expect it)
- no category configured behaves exactly as before

**Verified in both directions.** With the source change reverted and the tests kept, the
new test fails on a clean tree with the concrete path diff:

```
AssertionError: assert '.../nzbget/Saga.001' == '.../nzbget/Comics/Saga.001'
```

Full suite on this branch: **2830 passed, 1 skipped**, versus **2827 passed, 1 skipped**
on `main` — 3 added, none modified.

## Live verification

Deployed on a running instance:

```
NZBGet category subdirectory in use. Directory set to :
  /azuredata1/usenet/completed/Comics/One-Punch.Man.v30.2025.Digital.LuCaZ
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed NZBGet post-processing for downloads stored in configured category subdirectories.
  - Comicarr now locates completed downloads under the category folder when the direct download directory is unavailable.
  - Prevented affected post-processing tasks from stopping prematurely and blocking other Folder Monitor activity.
  - Preserved the existing direct-folder behavior when that path is available or no category is configured.

### DIFF
--- a/comicarr/app/downloads/postprocess_pipeline.py
+++ b/comicarr/app/downloads/postprocess_pipeline.py
@@ -55,6 +55,7 @@ class PostProcessInputContext:
     sab_direct_unpack: bool
     sab_directory: str | None
     nzbget_directory: str | None
+    nzbget_category: str | None = None
 
 
 @dataclass(frozen=True)
@@ -122,6 +123,13 @@ class PostProcessInputStage:
             elif context.nzbget_directory not in (None, "None"):
                 self._log.fdebug(f"{context.module} NZB name as passed from NZBGet: {context.nzb_name}")
                 folder = os.path.join(context.nzbget_directory, context.nzb_name)
+                if not self._path_exists(folder) and context.nzbget_category not in (None, "", "None"):
+                    categorised = os.path.join(context.nzbget_directory, context.nzbget_category, context.nzb_name)
+                    if self._path_exists(categorised):
+                        folder = categorised
+                        self._log.fdebug(
+                            f"{context.module} NZBGet category subdirectory in use. Directory set to : {folder}"
+                        )
                 self._log.fdebug(f"{context.module} NZBGET Download folder option enabled. Directory set to : {folder}")
 
         return PostProcessInputResult(folder)
--- a/.changeset/nzbget-category-subdirectory.md
+++ b/.changeset/nzbget-category-subdirectory.md
@@ -1,0 +1,5 @@
+---
+"comicarr": patch
+---
+
+NZBGet downloads that land under a configured category subdirectory are now found and post-processed. NZBGet files a completed download under `<DestDir>/<category>/`, but Comicarr only looked in `<DestDir>/`, so the folder it resolved did not exist. Post-processing aborted before reaching a terminal stage and the stranded obligation held the post-processing lock, which stopped Folder Monitor from sweeping anything else.
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -596,6 +596,7 @@ class PostProcessor(object):
                 sab_direct_unpack=(comicarr.CONFIG.SAB_DIRECT_UNPACK if use_sabnzbd else False),
                 sab_directory=(comicarr.CONFIG.SAB_DIRECTORY if use_sabnzbd else None),
                 nzbget_directory=(comicarr.CONFIG.NZBGET_DIRECTORY if use_nzbget else None),
+                nzbget_category=(comicarr.CONFIG.NZBGET_CATEGORY if use_nzbget else None),
             )
         )
         self.nzb_folder = input_result.folder
--- a/tests/unit/test_postprocess_input_resolution.py
+++ b/tests/unit/test_postprocess_input_resolution.py
@@ -184,3 +184,23 @@ def test_nzbget_resolution_without_category_keeps_plain_join(tmp_path):
 
     assert result == "manga"
     assert processor.nzb_folder == str(tmp_path / "nzbget" / "Saga.001")
+
+
+def test_nzbget_resolution_keeps_plain_join_when_neither_path_exists(tmp_path):
+    """With a category set but nothing on disk, the plain join must survive.
+
+    The fallback is a probe, not a rewrite: if the categorised path is not
+    there either, there is no evidence the category is in play, and the
+    historical path is the one whose failure the caller already reports. This
+    is the branch that pins the `_path_exists(categorised)` guard -- without
+    it the resolver silently redirects every unresolvable NZBGet download into
+    a category subdirectory that does not exist, and the "unable to locate"
+    error then names a path the operator never configured.
+    """
+    config = _config(tmp_path, NZBGET_CATEGORY="Comics")
+    processor, _queue = _processor(tmp_path / "incoming", config=config, use_nzbget=1)
+
+    result, _manga = _run(processor, config, use_nzbget=1)
+
+    assert result == "manga"
+    assert processor.nzb_folder == str(tmp_path / "nzbget" / "Saga.001")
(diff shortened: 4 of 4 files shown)


comicarr/app/downloads/postprocess_pipeline.py
```diff
@@ -122,6 +123,13 @@ def resolve(self, context: PostProcessInputContext) -> PostProcessInputResult:
             elif context.nzbget_directory not in (None, "None"):
                 self._log.fdebug(f"{context.module} NZB name as passed from NZBGet: {context.nzb_name}")
                 folder = os.path.join(context.nzbget_directory, context.nzb_name)
+                if not self._path_exists(folder) and context.nzbget_category not in (None, "", "None"):
```
frankieramirez, line 126: This needs a patch changeset or it skips CHANGELOG and What's New. CI only warns.

NZBGet downloads that land under a configured category subdirectory are now found and post-processed.

comicarr/app/downloads/postprocess_pipeline.py
```diff
@@ -122,6 +123,13 @@ def resolve(self, context: PostProcessInputContext) -> PostProcessInputResult:
...
+                    categorised = os.path.join(context.nzbget_directory, context.nzbget_category, context.nzb_name)
+                    if self._path_exists(categorised):
```
frankieramirez, line 128: You can delete this exists check and the three new tests still pass. They never take the false branch. Add a case with category set and both dirs missing, and assert nzb_folder stays the plain join.

comicarr/app/downloads/postprocess_pipeline.py
azuretek: No problem — added as `.changeset/nzbget-category-subdirectory.md` in e74f8e7, leading with your line and adding the operator-visible consequence:

> NZBGet downloads that land under a configured category subdirectory are now found and post-processed. NZBGet files a completed download under `<DestDir>/<category>/`, but Comicarr only looked in `<DestDir>/`, so the folder it resolved did not exist. Post-processing aborted before reaching a terminal stage and the stranded obligation held the post-processing lock, which stopped Folder Monitor from sweeping anything else.

Included the lock detail because that's the part an operator actually notices — the symptom isn't "one download didn't import", it's "nothing imports any more".

comicarr/app/downloads/postprocess_pipeline.py
azuretek: You're right, and I checked it rather than taking my word for it — deleting the `if self._path_exists(categorised):` guard entirely leaves **all 10 tests green**. The three category tests only ever cover the outer condition: one asserts the categorised hit, and the other two short-circuit before line 128 (plain path present, or category unset). The false branch was genuinely unpinned.

Added the case you asked for in e74f8e7 — `test_nzbget_resolution_keeps_plain_join_when_neither_path_exists`: category set, neither directory on disk, asserting `nzb_folder` stays the plain join. With the guard removed it fails as:

```
- p0/nzbget/Saga.001
+ p0/nzbget/Comics/Saga.001
```

Kept the guard rather than deleting it, and the test comment says why: the fallback is a probe, not a rewrite. If the categorised path isn't there either, there's no evidence the category is in play — and redirecting anyway would make the "unable to locate" error report a path the operator never configured, which is a worse thing to hand someone debugging this than the plain join they recognise.

Full suite green (2817 passed), `ruff check` / `ruff format --check` clean.

azuretek:
Added a changeset here too — `.changeset/nzbget-category-subdirectory.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

comicarr/app/downloads/postprocess_pipeline.py
frankieramirez: That's the right shape. Folder Monitor stalling is what operators actually see.

comicarr/app/downloads/postprocess_pipeline.py
frankieramirez: That's the pin. Keeping the guard as a probe is right; the new test would fail if it rewrote the path anyway.

frankieramirez (approved):

comicarr/app/downloads/postprocess_pipeline.py
```diff
@@ -122,6 +123,13 @@ def resolve(self, context: PostProcessInputContext) -> PostProcessInputResult:
...
+                        folder = categorised
+                        self._log.fdebug(
+                            f"{context.module} NZBGet category subdirectory in use. Directory set to : {folder}"
```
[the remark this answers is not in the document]
azuretek: Declining this one: `context.module` already carries its own brackets, so the suggested form would double them.

`PostProcessor.__init__` sets it at `comicarr/postprocessor.py:123-125`:

```python
self.module = module + "[POST-PROCESSING]"
...
self.module = "[POST-PROCESSING]"
```

So the line already emits `[POST-PROCESSING] NZBGet category subdirectory in use. ...`, which is the bracketed context the guideline asks for. Applying the suggestion would render `[[POST-PROCESSING]]`.

It would also make this the only line in the file shaped that way. Every sibling `fdebug` in `PostProcessInputResolver` interpolates it bare, including the two immediately around it (`postprocess_pipeline.py:78, 83-84, 87, 91, 99, 104, 117-118, 121, 124`).

PR #815, opened 2026-09-01
fix(search): snatch manga volume releases found by searchforissue

### LINKED ISSUE #811: Manga volume releases are found on the indexer but never snatched
## Summary

For a manga series tracked by volume, every search runs, the provider returns the correct
releases, and nothing is ever grabbed. There is no error — the searches simply complete
and the issues stay `Wanted`.

The releases are present and correctly named. They are discarded by five separate gates,
each of which produces the identical symptom, so fixing one only makes the next visible.

## Environment

- Based on `main` @ `01fbeb5`
- Series with `ContentType = manga`, issues carrying `VolumeNumber`
- NZBGeek (newznab); releases named `One-Punch Man v01 (2014) (Digital) (F) (BlackManta-Empire)`

## The RSS path already does this correctly

This is the clearest evidence that the behaviour below is unintended rather than
unimplemented. `comicarr/rsscheck.py:1762` calls `search_init()` and passes the full
manga contract:

```python
search.search_init(
    ...
    booktype="manga",
    content_type="manga",
    chapter_number=None if chapter_number is None else str(chapter_number),
    volume_number=None if volume_number is None else str(volume_number),
)
```

`searchforissue()` — the path behind the API, search-missing, and the scheduled backlog —
calls the same function at `comicarr/search.py:2577` and passes `content_type` but
**omits both numbers**:

```python
foundNZB, prov = search_init(
    ...
    booktype=booktype,
    ignore_booktype=ignore_booktype,
    content_type=content_type,
)
```

So the manga contract is wired in one caller and partially wired in the other.

## The five gates

| # | Gate | Effect |
|---|---|---|
| 1 | `searchforissue()` omits `chapter_number` / `volume_number` | `_build_manga_search_terms()` receives `(None, None)` and returns `[]`. No `vNN` term is ever built, and no `[SEARCH-MANGA]` line is logged, so there is nothing to notice. |
| 2 | A volume target still gets the issue number appended | The query goes out as `<series> v01 001`, which matches nothing. A bare-volume query shape already exists in the same loop but is gated on `booktype in ("TPB","HC","GN")` **and** `IssueNumber is None`; a manga volume target has an issue number, so it never reaches it. |
| 3 | Results are matched against the injected search term | `_build_manga_search_terms` injects `<series> v01` into `AlternateSearch`, and `gen_altnames()` yields it as a `ComicName`. The matcher then compares the release's parsed name `<series>` against `<series> v01` — a guaranteed mismatch. |
| 4 | No acceptance arm for a release with no issue number | The volume comparison succeeds and is logged, then no arm consumes it. The existing arms cover TPB double-pass and One-Shot only. |
| 5 | `vNN` is compared against `ComicVersion` | For a periodical, `vNN` means *which run of the series*; for manga it means *which book*. The comic reading is applied, so volume 30 is compared against the series' `ComicVersion` of 1 and rejected. |

## Observed

Gate 1 — every query issued in issue form, with no volume term at all:

```
Sending request to RSS for One-Punch Man : 001 (2015)
```

Gate 5 — the release is found and parsed correctly, then discarded on the version
comparison:

```
checking search result: One-Punch Man v30 (2025) (Digital) (LuCaZ)
FCVersion: 30
DCVersion: 1
Versions wrong. Ignoring possible match.
```

**Gate 5 makes the bug look intermittent.** `ComicVersion` defaults to `1`, so volume 1
alone passes this gate by coincidence while every other volume is discarded. That reads
as "manga mostly doesn't work" rather than as a systematic rejection.

## Reproduce

1. Add a series and set `ContentType = manga`.
2. Populate `VolumeNumber` on its issue rows.
3. Mark a volume `Wanted` and search it.
4. Observe the query in the log: issue form only, no `vNN` term.
5. Confirm on the indexer that a matching `<series> vNN` release exists.

## Proposed fix

Pass the numbers from `searchforissue()` as `rsscheck.py` already does, then follow the
consequences: search a volume target without the issue-number suffix, match results
against the real series name rather than the injected term, add an acceptance arm for a
volume release, and read `vNN` as the volume for manga rather than as the series run.

Each step is narrowly gated on a manga volume pass, so comic behaviour is untouched.

azuretek:
Fixes #811.

## Problem

A manga series tracked by volume runs every search, receives the correct releases from
the provider, and grabs none of them. Five gates each discard the release, producing the
identical symptom, so fixing one only exposes the next.

The RSS path already implements the manga contract correctly at `rsscheck.py:1762`.
`searchforissue()` calls the same `search_init()` and passes `content_type` but omits
`chapter_number` and `volume_number`, so `_build_manga_search_terms()` always receives
`(None, None)` and returns `[]`.

## Change

Six commits, one per gate plus one refactor:

| Commit | Gate |
|---|---|
| `pass manga chapter/volume numbers from searchforissue` | 1 — the numbers reach `search_init()`, as the RSS caller already does |
| `search manga volume targets without an issue-number suffix` | 2 — a volume target queries `<series> v01`, not `<series> v01 001` |
| `match manga volume results against the real series name` | 3 — carry the match name alongside the term so results compare against `<series>` |
| `accept a manga volume release on its volume number` | 4 — an acceptance arm for a release that carries no issue number |
| `give the volume rules one owner in the ledger` | refactor — see below |
| `read a manga release's vNN as the volume, not the series run` | 5 — a manga arm ahead of the `ComicVersion` comparison |

Each is narrowly gated on a manga volume pass; a comic search cannot reach any of them.
`manga_volume_satisfies()` returns `False` on missing or non-numeric data rather than
defaulting true, so a parse failure can never snatch an arbitrary volume.

## The refactor, and one intentional behaviour change

While adding gate 4 I found the same two facts already implemented in
`app/downloads/service.py::_pack_row_matches` — "is this target a volume?" and "do these
two volume numbers match?". Rather than leave a second copy, both now live in
`app/manga/ledger.py` beside `normalize_volume_number()`, which already owned what a
volume number means, and `_pack_row_matches` delegates to them.

**This changes two fractional-volume cases, and I would rather flag it than have a
reviewer find it:**

| Case | Before | After |
|---|---|---|
| `1.5` vs `1` | `True` (`int(float(...))` truncated) | `False` |
| `1.5` vs `1.5` | `False` (`int("1.5")` raised) | `True` |

Both look like corrections to me — `normalize_volume_number` already models fractional
volumes via `format(number, "g")`, so the truncating comparison was the odd one out.
Every integer case is unchanged, which is why the existing pack tests pass untouched. If
you would rather preserve the old comparison exactly, I am happy to keep the shared
helper and restore truncation.

## Test

Eleven tests added across `tests/unit/test_manga_ledger.py` and
`tests/unit/test_manga_live_paths.py`. The comparison cases live once, in the ledger
tests; the acceptance test only pins that acceptance defers to them, so the two cannot
drift.

Full suite on this branch: **2838 passed, 1 skipped**, versus **2827 passed, 1 skipped**
on `main` — 11 added, none modified.

**On asserting wiring by AST.** `searchforissue()` is ~900 lines behind provider config
and a global search lock, so invoking it in a unit test is impractical — but gate 1 *is*
a wiring omission, so the test parses the call site and asserts the keywords are passed.
The repo already uses this idiom in `test_init_arms_manga_sync_after_pause`. The same
assertion run against a clean `main` reports:

```
searchforissue -> search_init call sites: 1
  passes content_type: True | chapter_number: False | volume_number: False
```

## Live verification

Deployed on a running instance. Before, every query was issue form; after, the volume
term is generated and the release is snatched:

```
before:  Sending request to RSS for One-Punch Man : 001 (2015)
after:   I'm looking for One-Punch Man v01 issue: 1 (2015) using NZBGeek (newznab)
         [MANGA] volume v30 matches the volume searched for
         [UPDATER] Updated the status (Snatched) complete for One-Punch Man Issue: 1
```

## Summary by CodeRabbit

* **New Features**
  * Added manga volume-aware searching and release matching.
  * Volume searches now support alternate titles and volume-only queries.
  * Fractional volumes, such as `v01.5`, are preserved and matched accurately.

* **Bug Fixes**
  * Improved volume detection and comparison across common naming formats.
  * Prevented chapters from being incorrectly treated as volumes.
  * Improved matching for volume releases without issue numbers.

### DIFF
--- a/comicarr/app/downloads/service.py
+++ b/comicarr/app/downloads/service.py
@@ -30,6 +30,7 @@ from comicarr.app.downloads import queries as dl_queries
 from comicarr.app.downloads.completed_path import resolve_completed_download_file
 from comicarr.app.downloads.ddl_commands import DDLCommand, DDLCommandError
 from comicarr.app.downloads.pp_commands import PostProcessCommandError, configured_roots, validate_postprocess_item
+from comicarr.app.manga.ledger import is_volume_target, volume_numbers_match
 from comicarr.downloaders import mediafire, mega, pixeldrain
 from comicarr.tables import annuals, comics, ddl_info, issues, storyarcs, weekly
 
@@ -983,13 +984,10 @@ def _pack_row_matches(row, int_iss, iss_item, kind):
     chapter = row.get("ChapterNumber")
     if kind == "volume":
         if volume not in (None, ""):
-            try:
-                return int(float(volume)) == int(iss_item)
-            except (TypeError, ValueError):
-                return False
+            return volume_numbers_match(volume, iss_item)
         if chapter not in (None, ""):
             return False
-    elif volume not in (None, "") and chapter in (None, ""):
+    elif is_volume_target(chapter, volume):
         return False
     return row["Int_IssueNumber"] == int_iss
 
--- a/comicarr/app/manga/acquisition.py
+++ b/comicarr/app/manga/acquisition.py
@@ -154,10 +154,17 @@ def search_plan_for_series(series, issues):
 
 
 def _pad_volume(number):
+    # Keeps the fraction, the same way _pad_chapter does. Half volumes are
+    # real ("v01.5"), and truncating one to "v01" searches the wrong volume --
+    # the exact volume comparison in _pack_row_matches then rejects the 1 it
+    # got against the 1.5 it wanted, so the row can never snatch.
     try:
-        return "%02d" % int(float(number))
+        value = float(number)
     except (TypeError, ValueError):
         return str(number) if number not in (None, "") else None
+    if value == int(value):
+        return "%02d" % int(value)
+    return "%02d.%s" % (int(value), str(round(value % 1, 1))[2:])
 
 
 def _pad_chapter(number):
--- a/comicarr/app/manga/acquisition.py
+++ b/comicarr/app/manga/acquisition.py
@@ -9,6 +9,8 @@
 
 """What automatic search should enqueue for a manga Series."""
 
+from decimal import Decimal, InvalidOperation
+
 from comicarr.app.manga.ledger import last_released_volume, normalize_volume_number
 
 MONITOR_MODES = frozenset({"blended", "volumes", "chapters"})
@@ -153,28 +155,42 @@ def search_plan_for_series(series, issues):
     )
 
 
-def _pad_volume(number):
-    # Keeps the fraction, the same way _pad_chapter does. Half volumes are
-    # real ("v01.5"), and truncating one to "v01" searches the wrong volume --
-    # the exact volume comparison in _pack_row_matches then rejects the 1 it
-    # got against the 1.5 it wanted, so the row can never snatch.
+def _pad_number(number, width):
+    """Zero-pad a volume or chapter number, keeping its fraction EXACTLY.
+
+    Half instalments are real ("v01.5", "c001.5"), and truncating one searches
+    the wrong book -- the exact comparison in _pack_row_matches then rejects
+    the 1 it got against the 1.5 it wanted, so the row can never snatch.
+
+    The fraction is carried as text rather than through float arithmetic.
+    `round(value % 1, 1)` keeps only ONE fractional digit, so it silently
+    rewrote the number it was meant to preserve: 1.25 searched "v01.2" and
+    1.75 searched "v01.8" -- not a truncation of the wanted volume but a
+    DIFFERENT one, which the same exact comparison then rejects.
+
+    A non-numeric value is returned as given: some series number instalments
+    in ways no padding rule can improve on, and inventing one would be worse
+    than leaving it alone.
+    """
+    if number in (None, ""):
+        return None
+    text = str(number).strip()
     try:
-        value = float(number)
-    except (TypeError, ValueError):
-        return str(number) if number not in (None, "") else None
-    if value == int(value):
-        return "%02d" % int(value)
-    return "%02d.%s" % (int(value), str(round(value % 1, 1))[2:])
+        value = Decimal(text)
+    except (InvalidOperation, TypeError, ValueError):
+        return text
+    if value == value.to_integral_value():
+        return "%0*d" % (width, int(value))
+    whole, _, fraction = format(value, "f").partition(".")
+    return "%0*d.%s" % (width, int(whole), fraction)
+
+
+def _pad_volume(number):
+    return _pad_number(number, 2)
 
 
 def _pad_chapter(number):
-    try:
-        value = float(number)
-    except (TypeError, ValueError):
-        return str(number) if number not in (None, "") else None
-    if value == int(value):
-        return "%03d" % int(value)
-    return "%03d.%s" % (int(value), str(round(value % 1, 1))[2:])
+    return _pad_number(number, 3)
 
 
 def _as_float(value):
--- a/comicarr/app/manga/ledger.py
+++ b/comicarr/app/manga/ledger.py
@@ -15,6 +15,7 @@ Owning a volume covers only the chapters that mapping actually lists.
 """
 
 import re
+from decimal import Decimal, InvalidOperation
 
 from comicarr.app.acquisition.models import Fulfillment
 
@@ -64,11 +65,31 @@ def volume_numbers_match(left, right):
     Returns False whenever either side has no usable volume, so a parse
     failure can never be read as a match and claim an arbitrary volume.
     """
-    left_value = normalize_volume_number(_strip_volume_marker(left))
-    right_value = normalize_volume_number(_strip_volume_marker(right))
+    left_value = _exact_volume(_strip_volume_marker(left))
+    right_value = _exact_volume(_strip_volume_marker(right))
     if left_value is None or right_value is None:
         return False
     return left_value == right_value
+
+
+def _exact_volume(value):
+    """A volume reference as an exact number, or None when it is not one.
+
+    normalize_volume_number is a DISPLAY normaliser: it preserves arbitrary
+    text and formats floats to six significant digits. Comparing through it
+    therefore made "TPB" equal "TPB" -- two releases with no volume at all
+    reading as the same volume -- and "1.0000001" equal "1". Neither is a
+    volume identity, and the contract above says so.
+    """
+    if value is None:
+        return None
+    text = str(value).strip()
+    if not text:
+        return None
+    try:
+        return Decimal(text).normalize()
+    except (InvalidOperation, TypeError, ValueError):
+        return None
 
 
 def _strip_volume_marker(value):
--- a/comicarr/app/manga/ledger.py
+++ b/comicarr/app/manga/ledger.py
@@ -14,10 +14,15 @@ chapters and volume→chapter containment come from unfiltered `/aggregate`.
 Owning a volume covers only the chapters that mapping actually lists.
 """
 
+import re
+
 from comicarr.app.acquisition.models import Fulfillment
 
 _OPERATOR_FIELDS = ("Status", "AcquisitionIntent", "Location", "status", "acquisitionIntent", "location")
 _NON_VOLUME = frozenset({"", "none", "null", "unknown"})
+# Only strips a marker that actually introduces a number, so "vTPB" is left
+# alone and never compares equal to a real volume.
+_VOLUME_MARKER = re.compile(r"^v(?:ol(?:ume)?)?\.?\s*(?=\d)", re.IGNORECASE)
 
 
 def normalize_volume_number(value):
@@ -34,6 +39,42 @@ def normalize_volume_number(value):
     if number.is_integer():
         return str(int(number))
     return format(number, "g")
+
+
+def is_volume_target(chapter_number, volume_number):
+    """Is this pair a volume rather than a chapter?
+
+    The single owner of the volume-vs-chapter rule: a volume has a volume
+    number and no chapter number. Stated once because getting it wrong in
+    either direction is silently destructive — a chapter treated as a volume
+    lets a pack claim "chapter 7" as "volume 7", and a volume treated as a
+    chapter is searched as ``c001`` and never matches a ``v01`` release.
+    """
+    return volume_number not in (None, "") and chapter_number in (None, "")
+
+
+def volume_numbers_match(left, right):
+    """Do two volume references denote the same volume?
+
+    The same volume is written differently everywhere it appears: a release
+    says ``v01``, the ledger stores ``1``, a pack range yields the integer 1.
+    Both sides are stripped of a leading volume marker and canonicalised
+    through :func:`normalize_volume_number` before comparison.
+
+    Returns False whenever either side has no usable volume, so a parse
+    failure can never be read as a match and claim an arbitrary volume.
+    """
+    left_value = normalize_volume_number(_strip_volume_marker(left))
+    right_value = normalize_volume_number(_strip_volume_marker(right))
+    if left_value is None or right_value is None:
+        return False
+    return left_value == right_value
+
+
+def _strip_volume_marker(value):
+    if value is None:
+        return None
+    return _VOLUME_MARKER.sub("", str(value).strip())
 
 
 def chapter_id(comic_id, chapter_number):
--- a/comicarr/search.py
+++ b/comicarr/search.py
@@ -2567,9 +2567,7 @@ def searchforissue(
                     content_type = "manga" if series_kind.is_manga(comic) else "comic"
                     if content_type == "manga":
                         manga_row = db.select_one(
-                            select(issues.c.ChapterNumber, issues.c.VolumeNumber).where(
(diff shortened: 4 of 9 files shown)


azuretek:

comicarr/search.py
```diff
@@ -326,6 +327,7 @@ def search_init(
                 AlternateSearch = manga_alt_str + "##" + AlternateSearch
             else:
                 AlternateSearch = manga_alt_str
+            manga_volume_terms = manga_volume_search_terms(ComicName, chapter_number, volume_number)
```
frankieramirez: A same-numbered chapter release can snatch this volume row. gen_altnames still puts the unsuffixed series name first, and search_the_matrix stops on the first hit, so One-Punch Man 030 matches the issue arm and the v30 pass never runs. Search the unsplit volume terms first, and skip the issue-number pass when manga_volume_terms is set.

comicarr/search.py
```diff
@@ -1092,8 +1096,14 @@ def NZB_SEARCH(
     foundc["lastrun"] = provider_stat["lastrun"]
     done = False
 
+    # Set only when this pass is searching a manga volume term. Carries the
+    # real series name, because the term itself ("<series> v01") is a query
+    # rather than a name and would fail every series comparison.
+    manga_match_name = (manga_volume_terms or {}).get(ComicName)
```
frankieramirez, line 1122: Gantz! and Working!! never hit this lookup. Volume terms go through gen_altnames, which splits on ! (and treats !! as a priority alt), so keys like Gantz! v01 never show up as ComicName. Append the unsplit volume terms after gen_altnames builds the name list so this .get actually fires.

comicarr/app/downloads/service.py
```diff
@@ -983,13 +984,10 @@ def _pack_row_matches(row, int_iss, iss_item, kind):
     chapter = row.get("ChapterNumber")
     if kind == "volume":
         if volume not in (None, ""):
-            try:
-                return int(float(volume)) == int(iss_item)
-            except (TypeError, ValueError):
-                return False
+            return volume_numbers_match(volume, iss_item)
```
frankieramirez, line 987: A 1.5 volume now fails to snatch. _pad_volume still queries v01 and FileChecker captures only the integer, then this rejects 1 vs 1.5. Keep the exact match, and make _pad_volume plus the manga volume regex keep the fraction the way _pad_chapter already does.

comicarr/search.py
azuretek: No problem — and you were right on both the mechanism and the consequence. Fixed in ff44ff7.

Your diagnosis held exactly: `gen_altnames()` appends the unsuffixed name at position 2 (or 1, when there's no `!!` priority alt), the volume terms land after it as ordinary `##` alternates, and `search_the_matrix` returns on the first hit — so the v30 pass was unreachable whenever a chapter release numbered 030 existed. Worse than a miss, since the row then reads Downloaded with a chapter file in it.

Rather than reorder inside `AlternateSearch`, I took the volume terms out of it entirely — that string is the thing doing the damage in your other comment too. New `manga_volume_altnames()` takes the *finished* name list from `gen_altnames()` and rewrites each entry into its volume query:

```python
entries, terms = manga_volume_altnames(altnames, volume_number)
```

So there is no bare name left in the list for `NZB_SEARCH` to append an issue number to — the issue-number pass is skipped structurally rather than by a flag check. Applied in both the rss and api loops.

Two things I did beyond the letter of the suggestion, both to avoid a regression:

- **Alternates keep their coverage.** Rather than dropping to just the primary term, each alternate title gets its own `"<alt> vNN"`. A manga with an English alt title is still searchable by volume.
- **It falls back.** If no term can be built (an unusable volume number), it returns the original list untouched, so a series is never left unsearchable by this path.

Tests: `test_manga_volume_altnames_leaves_no_bare_name_to_search_by_issue_number` asserts the bare name is gone and the alternates survived, and `test_search_init_wires_volume_queries_instead_of_alternatesearch` pins that both loops rewrite the list and that the `AlternateSearch` injection is reachable only for a non-volume target.

comicarr/search.py
azuretek: Confirmed and fixed in ff44ff7 — and I wrote a characterisation test for it, because the failure is invisible otherwise:

```python
def test_gen_altnames_shreds_a_volume_term_carried_in_alternatesearch():
    names = [x["ComicName"] for x in search.gen_altnames("Gantz!", "Gantz! v01", None, "want")]
    assert "Gantz! v01" not in names
```

That passes today. The split is `re.split(r"[\!\!]+|[\#\#]+", AlternateSearch)` — a character class, so it's really "split on any `!` or `#`" — plus the `!!` alt-priority scan ahead of it. Either way `Gantz! v01` never arrives intact, so `manga_volume_terms.get(ComicName)` missed and the pass quietly fell back to an issue-number search. Silent, which is the bad part.

The fix is the same structural one as your other comment: volume terms no longer ride `AlternateSearch` at all. `manga_volume_altnames()` builds them from the finished name list *after* `gen_altnames()` has done its splitting, so nothing can shred them:

```python
entries, terms = manga_volume_altnames(gen_altnames("Gantz!", None, None, "want"), "1")
# entries -> ["Gantz! v01"], terms -> {"Gantz! v01": "Gantz!"}
```

`test_manga_volume_altnames_keeps_a_bang_title_unsplit` covers it. The `terms` mapping still carries the real series name for the matcher, since `"Gantz! v01"` is a query and the release parses as `"Gantz!"`.

comicarr/app/downloads/service.py
azuretek: Good catch — kept the exact match and fixed both producers, as you suggested. ff44ff7.

The asymmetry was right there and I'd missed it: `_pad_chapter` already handles fractions and `_pad_volume` didn't, and the same split runs through FileChecker's two patterns.

```python
# _pad_volume  -> int(float(number))            "1.5" -> "01"
# _pad_chapter -> keeps the fraction            "1.5" -> "001.5"

# chapter pattern: (\d+(?:\.\d+)?)   keeps it
# volume pattern:  (\d+)             drops it
```

Both now match their chapter counterpart, so `_pad_volume("1.5")` → `"01.5"` and the volume pattern captures `01.5`. Whole volumes are unchanged (`"2"` → `"02"`).

Keeping the exact comparison was the right call — the old `int(float(volume)) == int(iss_item)` didn't just tolerate 1.5, it actively mismatched: volume 1 and volume 1.5 both truncate to 1, so either could claim the other's release.

Tests: `test_half_volume_keeps_its_fraction_through_search_and_match` asserts the term is `"Kanojo v01.5"` and that `volume_numbers_match("v01", "1.5")` is still **False**, and `test_filechecker_captures_a_fractional_manga_volume` covers the pattern — including that `"Kanojo v01.cbz"` still reports `01`, so an extension can't be read as a fraction.

Both fail on the pre-fix code (`'Kanojo v01' != 'Kanojo v01.5'`).

azuretek:
Added a changeset here too — `.changeset/manga-volume-search.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

frankieramirez (dismissed):

comicarr/app/manga/acquisition.py
```diff
@@ -154,10 +154,17 @@ def search_plan_for_series(series, issues):
...
+    # the exact volume comparison in _pack_row_matches then rejects the 1 it
+    # got against the 1.5 it wanted, so the row can never snatch.
     try:
-        return "%02d" % int(float(number))
+        value = float(number)
     except (TypeError, ValueError):
         return str(number) if number not in (None, "") else None
+    if value == int(value):
+        return "%02d" % int(value)
+    return "%02d.%s" % (int(value), str(round(value % 1, 1))[2:])
```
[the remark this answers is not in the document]
azuretek: Valid, and worse than the report says — the same construct is in `_pad_chapter`, so chapters had it too. Fixed in 72fd3ff.

Confirmed before changing anything:

```
'1.25' -> vol='01.2'  chap='001.2'
'1.75' -> vol='01.8'  chap='001.8'
'1.05' -> vol='01.1'  chap='001.1'
```

`1.75 -> 01.8` is the sharp one: not a truncation of the wanted volume but a *different* one, which the exact comparison then rejects — so the row can never snatch. That is exactly the failure the half-volume work on this branch was written to fix, still live for anything past one decimal place.

The fraction is carried as text now, through `Decimal`, and the two functions share one implementation rather than repeating the arithmetic:

```python
whole, _, fraction = format(value, "f").partition(".")
return "%0*d.%s" % (width, int(whole), fraction)
```

`1.5 -> 01.5` and `2 -> 02` are unchanged. Restoring the rounding fails the new tests.

comicarr/app/manga/ledger.py
```diff
@@ -36,6 +41,42 @@ def normalize_volume_number(value):
...
+    The same volume is written differently everywhere it appears: a release
+    says ``v01``, the ledger stores ``1``, a pack range yields the integer 1.
+    Both sides are stripped of a leading volume marker and canonicalised
+    through :func:`normalize_volume_number` before comparison.
+
+    Returns False whenever either side has no usable volume, so a parse
+    failure can never be read as a match and claim an arbitrary volume.
+    """
+    left_value = normalize_volume_number(_strip_volume_marker(left))
+    right_value = normalize_volume_number(_strip_volume_marker(right))
```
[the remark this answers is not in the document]
azuretek: Valid on both counts, and confirmed against the code before changing it:

```
volume_numbers_match('TPB', 'TPB')      -> True
volume_numbers_match('1.0000001', '1')  -> True
```

The function's own docstring says it "Returns False whenever either side has no usable volume, so a parse failure can never be read as a match and claim an arbitrary volume" — so the implementation contradicted its stated contract. `normalize_volume_number` is a *display* normaliser: it preserves arbitrary text and formats floats to six significant digits, which is right for showing a number and wrong for deciding identity.

Fixed in 72fd3ff with a comparison-specific parse:

```python
def _exact_volume(value):
    ...
    try:
        return Decimal(text).normalize()
    except (InvalidOperation, TypeError, ValueError):
        return None
```

`v01`/`01`/`1` still match, `v01.5` still matches `1.5`, and both regression cases you asked for are covered. Full suite is green, so nothing was relying on the text-equality behaviour.

comicarr/search_filer.py
```diff
@@ -871,7 +892,20 @@ def _match_entry(self, entry, is_info):
             logger.fdebug("SCVersion: %s" % S_ComicVersion)
             logger.fdebug("ComicYear: %s" % ComicYear)
 
-            if all(
+            if manga_volume_pass and manga_volume_satisfies(fndcomicversion, findcomiciss):
```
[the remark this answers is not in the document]
azuretek: Valid. Traced it: the version arms only set `fndcomicversion` for an all-digit `vNN` — `"v01.5"[1:]` is `"01.5"`, which is not `.isdigit()`, so every arm falls through and it stays `None`. FileChecker parsed the volume perfectly well; the matcher just could not see it.

That makes it the same bug as the padding one in a second place: this branch teaches the search to *ask* for `v01.5` and then rejects the result.

Fixed in 72fd3ff by falling back to the parsed volume:

```python
parsed_volume = fndcomicversion or parsed_comic.get("series_volume")
if manga_volume_pass and manga_volume_satisfies(parsed_volume, findcomiciss):
```

Kept `fndcomicversion` first so the whole-number path is unchanged, and scoped to this branch so periodical version handling is untouched.

comicarr/search.py
```diff
@@ -4235,12 +4288,71 @@ def _build_manga_search_terms(series_name, chapter_num, volume_num):
...
+
+    entries = []
+    terms = {}
+    for entry in altnames or ():
+        name = entry.get("unaltered_ComicName") or entry.get("ComicName")
+        for term in search_terms_for_target(name, {"kind": "volume", "number": volume_num}):
+            if term in terms:
+                continue
+            terms[term] = name
+            entries.append({"ComicName": term, "unaltered_ComicName": term})
```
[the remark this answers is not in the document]
azuretek: Real, though it predates this branch — the query is built by hand rather than through `params`:

```python
cm = re.sub(r"[\&\:\?\,]", "", str(cm))
cm = re.sub(" ", "%20", str(cm))
```

`&`, `:`, `?` and `,` are stripped; `#` was not. So a `#` in a series name terminates the URL at the fragment and the `&apikey=` appended afterwards never reaches the provider — a silent empty result set rather than an error.

Fixed in 72fd3ff by adding `#` to that same strip list, which keeps one convention rather than introducing a second escaping scheme next to it.

Not moving the whole query to `params` here: `findurl` is assembled across several branches with hand-encoded values (`%20`, `%27`) already baked in, so converting it is a real change to the provider request path and wants its own PR and its own testing against live indexers. Happy to open that separately if it is wanted.

azuretek:

The replies on this PR cite `72fd3ff`, which was the head before the rebase. That commit is now `d6ac6f8`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/manga-search-20260903`.

PR #816, opened 2026-09-01
fix(manga): mark a placed manga volume file against its issue row

### LINKED ISSUE #812: Manga volume files are placed correctly but never marked Downloaded (Have stays 0)
## Summary

Manga volume files are moved into the series folder successfully, and are readable by a
downstream reader. Comicarr then reports `0 files` and the `Have` count never rises,
because the placed file is not matched back to any issue row.

Three independent causes, all sharing the same root: a manga volume file is identified by
its **volume**, and several code paths identify a file by its **issue number** only.

## Environment

- Based on `main` @ `01fbeb5`
- Series with `ContentType = manga`, issues carrying `VolumeNumber`
- Files named `One-Punch Man v20 (2020) (Digital) (LuCaZ).cbz`

## Observed

Post-processing places the file, then matches nothing:

```
Placed manga file: One-Punch Man v20 (2020) (Digital) (LuCaZ).cbz -> .../One-Punch Man (2014)
Unable to match ... to any chapter for One-Punch Man
Manga post-processing complete for One-Punch Man: 0 files
```

The folder scan finds the series, then selects nothing:

```
[FILENAME]: One-Punch Man v01 (2014) (Digital) (F) (BlackManta-Empire).cbr
Volume detected as : v01
[SCAN SUMMARY] ...cbr: candidates=1, selected=none
```

## Cause 1 — the parser anchors the number at the end of the stem

Every pattern in `comicarr/manga_parser.py` requires the volume or chapter token to end
the stem:

```python
_PAT_VOLUME_ONLY = re.compile(r"^(?P<series>.+?)\s+[Vv](?P<volume>\d+)\s*$")
```

That parses `One-Punch Man v20.cbz` but not
`One-Punch Man v20 (2020) (Digital) (LuCaZ).cbz`, because scene releases append year,
format and group tags after the volume token. `parse_manga_filename()` therefore returns
`None`.

Note the consequence: `_process_manga` **already has a correct volume-matching branch**,
and `rescan_manga_series` has one too. Neither is reachable, because `parsed` is `None`
before either branch runs.

The same `for pattern in _PATTERNS` loop is duplicated in
`parse_manga_chapter_number()`, so the blindness exists in both functions.

## Cause 2 — the folder scan hardcodes a collected-edition type list

The folder scan decides how to number a file inline:

```python
if (Type in ("TPB","HC","GN") and Total > 1) or (Type == "One-Shot" and Total == 1):
    just_the_digits = re.sub("[^0-9]", "", watchmatch["series_volume"])   # by volume
else:
    just_the_digits = watchmatch["justthedigits"]                          # by issue
```

Manga is not in that set, so it falls to the issue branch — and a manga volume release
carries no issue number, so `justthedigits` is `None`. The scan then compares every issue
in the series against nothing, which is the `candidates=1, selected=none` above.

**That same hardcoded tuple is tested in seven places** in `comicarr/postprocessor.py` —
three number derivations plus four guards that exempt collected editions from checks
which assume an issue number (the weekly-pull cross-check, the wrong-match rejection, and
two default-to-1 fallbacks). Every copy omits manga, so correcting one copy only moves
the failure to the next.

Verified against a clean `main` checkout:

```
UPSTREAM postprocessor.py
  hardcoded '"TPB"' occurrences: 7
```

## Cause 3 — the year gate rejects a matched volume

Once the volume matches, the file is still rejected on year:

```
[ISSUE-VERIFY][SeriesYear-Volume MATCH] Volume label ... matched to volume label of v01
[ISSUE-VERIFY][Issue Year FAILURE] Issue Year of 2015 does NOT match the year found
                                   in the filename of : 2014
```

The metadata provider dates the licensed English printing `2015`; the release is labelled
`(2014)`, the volume's original year. Manga volumes routinely differ this way, because
the two dates describe different printings. The existing rescue arm cannot help — it
requires `issue_year is None`, so it only saves files with no year in the name.

## Proposed fix

1. Retry the parse against the stem with trailing tag groups stripped, via a single
   shared helper used by both parser entry points. Try the full stem first so existing
   names that carry captures inside tags are unaffected.
2. Replace the seven hardcoded type tests with named predicates over one tuple, and add a
   manga arm using the existing `series_kind.is_manga()` rather than testing
   `ContentType` directly.
3. For manga only, treat a matched volume as settling the year: a volume number does not
   repeat across a series, so once it matches, the year cannot indicate a wrong file. A
   collected edition's year mismatch can still mean the wrong edition, so comic behaviour
   is unchanged.

azuretek:
Fixes #812.

## Problem

Manga volume files are placed into the series folder correctly and are readable, but
Comicarr reports `0 files` and `Have` never rises. Three causes, all the same root: a
manga volume file is identified by its **volume**, and several paths identify a file by
its **issue number** only.

## Change

Three commits:

### 1. `parse volume/chapter tokens followed by release tags`

Every pattern anchors the number at the end of the stem (`\s*$`), so
`<series> v20 (2020) (Digital) (Group).cbz` does not parse and
`parse_manga_filename()` returns `None`.

Worth noting what this unblocks rather than adds: `_process_manga` and
`rescan_manga_series` **already have correct volume-matching branches**. They were simply
unreachable, because `parsed` was `None` before either branch ran.

Adds `strip_trailing_tags()` and a shared `_match_patterns()` helper that tries the full
stem first and retries the stripped stem only when nothing matched. Full-stem-first keeps
the change purely additive — `[Group] Series - c001 (v03) [HQ].cbz` still returns
`volume=3, quality='HQ'`, which stripping first would have eaten.

The identical `for pattern in _PATTERNS` loop existed in `parse_manga_chapter_number()`
too, so both entry points now route through the one helper rather than leaving the same
blindness in the second.

### 2 + 3. `number a watchlisted manga file by its volume during folder scan`, then `give the folder scan one owner for volume-numbered series`

The folder scan derived the identifying number inline from a hardcoded
`("TPB","HC","GN")` / `One-Shot` test, which manga is not in — so a volume file yielded
no number and matched no issue.

That tuple was tested in **seven** places: three number derivations plus four guards that
exempt collected editions from checks assuming an issue number. Every copy omitted manga,
so the first commit fixed one site and the failure simply moved to the next; the second
commit gives the fact one owner.

Two predicates over one tuple, because the sites need different questions:

| Predicate | Question | Sites |
|---|---|---|
| `numbered_by_volume()` | type alone — is this series exempt from issue-number checks? | the 4 guards |
| `volume_identifies_file()` | same, qualified by run length — is a file *located* by its volume? | the 3 derivations |

They deliberately disagree for a one-entry TPB (`exempt=True, locates=False`), which is
why they are separate — collapsing them into one would have silently changed TPB
handling.

The manga signal comes from the existing `series_kind.is_manga()`, which reconciles
`ContentType` with the provider prefix, rather than testing either directly. Rows built
without the flag read as comics, so nothing else changes.

The third commit also adds `volume_match_settles_year()`: for manga only, a matched
volume settles the year, because a volume number does not repeat across a series. A
collected edition's year mismatch can still mean the wrong edition, so comic behaviour is
untouched.

## Test

Thirty tests added across `tests/unit/test_manga_parser.py`,
`tests/unit/test_manga_postprocessor.py` and `tests/unit/test_manga_live_paths.py`.

Full suite on this branch: **2857 passed, 1 skipped**, versus **2827 passed, 1 skipped**
on `main` — 30 added, none modified.

**On proving these fail without the fix.** Reverting the source makes the manga test
modules fail at *import*, because they exercise helpers this PR introduces — that proves
the helpers are new, not that the fix works. The meaningful assertions are therefore the
AST wiring tests, which check that production code actually consults the predicates. Run
against a clean `main` checkout they report:

```
UPSTREAM postprocessor.py
  calls volume_identifies_file : False
  WatchValues carries IsManga  : False
  hardcoded '"TPB"' occurrences: 7
```

One of those tests asserts `source.count('"TPB"') == 1` directly, so it fails on `main`
with `assert 7 == 1` and guards the consolidation against future copies.

The parser commit has ordinary behavioural coverage: with its source reverted and its
tests kept, 6 of its 8 tests fail; the other 2 are guards asserting preserved behaviour
and correctly pass on both sides.

## Live verification

Deployed on a running instance. The same log file holds the before and after across two
builds of the same file:

```
before:  [Issue Year FAILURE] 2015 does NOT match ... 2014   -> selected=none
after:   [MANGA][VOLUME MATCH] Volume v01 matched, so the year in the
         filename (2014) does not decide this file            -> selected=71856
```

`Have` went 12 → 13 with no manual intervention, and a full reconcile marked 12 placed
files `Downloaded`, matching the reader's view of the folder exactly.

## Ordering

This PR is only reachable once a manga volume is actually snatched, so it is best read
after #815. The changes themselves are independent — this
branch applies cleanly to `main` on its own and its suite is green.

### DIFF
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -77,25 +77,64 @@ def log_scan_summary(module, filename, candidate_count, selected_items, annual_c
     )
 
 
+_COLLECTED_TYPES = ("TPB", "HC", "GN")
+
+
+def numbered_by_volume(watch_values):
+    """Return whether this series names its files by volume rather than issue.
+
+    Collected editions and One-Shots have always been named this way, which is
+    why the scan exempts them from every check that assumes a file carries an
+    issue number: the weekly-pull cross-check, the "no issue number" rejection,
+    and the fallback that defaults a missing number to 1.
+
+    Manga belongs with them. In a periodical ``vNN`` names *which run* of the
+    series a release belongs to; in manga it names *which book*. A manga volume
+    file therefore carries no issue number at all, so applying the periodical
+    checks to it rejects a perfectly good file.
+
+    Rows built without the manga flag -- story arcs and one-offs assembled from
+    tables that do not carry it -- read as comics, which is the prior behaviour.
+    """
+    if watch_values.get("IsManga"):
+        return True
+    return watch_values["Type"] in _COLLECTED_TYPES + ("One-Shot",)
+
+
+def volume_match_settles_year(watch_values, volume_matched):
+    """Return whether a matched volume makes the filename's year irrelevant.
+
+    The year check exists because periodical issue numbers repeat across runs,
+    so ``Batman 1`` needs a year to say *which* Batman 1. A manga volume number
+    does not repeat -- there is one volume 1 -- so once it matches, the year
+    cannot be evidence of a wrong file.
+
+    It is routinely a mismatch, too: a provider dates the licensed English
+    printing while the release is labelled with the volume's original year. For
+    One-Punch Man v01 those are 2015 and 2014, and the file was rejected for it.
+
+    Deliberately manga-only. A collected edition's year mismatch can still mean
+    the wrong edition, so periodical and TPB behaviour is untouched.
+    """
+    return bool(volume_matched) and bool(watch_values.get("IsManga"))
+
+
 def volume_identifies_file(watch_values):
-    """Return whether this series numbers its files by volume rather than issue.
+    """Return whether a scanned file is located by volume rather than issue.
 
-    Collected editions (TPB/HC/GN spanning more than one entry) and One-Shots
-    have always been numbered this way, so a scanned file is located by the
-    ``vNN`` token in its name rather than by an issue number.
+    This is :func:`numbered_by_volume` qualified by the run length, because a
+    collected edition only numbers *files* by volume when the run actually has
+    volumes to distinguish: TPB/HC/GN spanning more than one entry, or a
+    One-Shot standing alone.
 
-    Manga joins them for a different reason. In a periodical, ``vNN`` names
-    *which run* of the series a release belongs to; in manga it names *which
-    book*, so the volume is the file's identity exactly as an issue number is
-    elsewhere. A manga volume file carries no issue number at all, so without
-    this the scan derives no number for it, matches it against every issue in
-    the series, and selects none of them.
+    Manga is not qualified that way. Volume 1 of a one-volume manga is still
+    identified by its volume, so the run length says nothing useful here.
     """
     if watch_values.get("IsManga"):
         return True
     series_type = watch_values["Type"]
     total = watch_values["Total"]
-    return (series_type in ("TPB", "HC", "GN") and total > 1) or (series_type == "One-Shot" and total == 1)
+    return (series_type in _COLLECTED_TYPES and total > 1) or (series_type == "One-Shot" and total == 1)
 
 
 def summarize_scan_matches(normal_items, arc_items):
@@ -1384,27 +1423,13 @@ class PostProcessor(object):
                             temploc = just_the_digits.replace("_", " ")
                             temploc = re.sub("[\\#']", "", temploc)
                         else:
-                            if any(
-                                [
-                                    cs["WatchValues"]["Type"] == "TPB",
-                                    cs["WatchValues"]["Type"] == "GN",
-                                    cs["WatchValues"]["Type"] == "HC",
-                                    cs["WatchValues"]["Type"] == "One-Shot",
-                                ]
-                            ):
+                            if numbered_by_volume(cs["WatchValues"]):
                                 temploc = "1"
                             else:
                                 temploc = None
                         datematch = "False"
 
-                        if temploc is None and all(
-                            [
-                                cs["WatchValues"]["Type"] != "TPB",
-                                cs["WatchValues"]["Type"] != "GN",
-                                cs["WatchValues"]["Type"] != "HC",
-                                cs["WatchValues"]["Type"] != "One-Shot",
-                            ]
-                        ):
+                        if temploc is None and not numbered_by_volume(cs["WatchValues"]):
                             logger.info(
                                 "this should have an issue number to match to this particular series: %s"
                                 % cs["ComicID"]
@@ -1662,14 +1687,9 @@ class PostProcessor(object):
                                             alts = x["AS_Alt"]
                                     alt_listing = [True if x.lower() == dynamic_seriesname else False for x in alts]
 
-                                    if any([cs["DynamicName"] == dynamic_seriesname, alt_listing]) and all(
-                                        [
-                                            cs["WatchValues"]["Type"] != "TPB",
-                                            cs["WatchValues"]["Type"] != "GN",
-                                            cs["WatchValues"]["Type"] != "HC",
-                                            cs["WatchValues"]["Type"] != "One-Shot",
-                                        ]
-                                    ):
+                                    if any(
+                                        [cs["DynamicName"] == dynamic_seriesname, alt_listing]
+                                    ) and not numbered_by_volume(cs["WatchValues"]):
                                         logger.fdebug(
                                             "name match exact : %s - %s" % (cs["DynamicName"], dynamic_seriesname)
                                         )
@@ -1798,15 +1818,7 @@ class PostProcessor(object):
                                 else:
                                     logger.fdebug("not a match")
 
-                                if all(
-                                    [
-                                        second_check is False,
-                                        cs["WatchValues"]["Type"] != "TPB",
-                                        cs["WatchValues"]["Type"] != "GN",
-                                        cs["WatchValues"]["Type"] != "HC",
-                                        cs["WatchValues"]["Type"] != "One-Shot",
-                                    ]
-                                ):
+                                if second_check is False and not numbered_by_volume(cs["WatchValues"]):
                                     logger.fdebug(
                                         "%s %s in filename don't match up to what's in the dB for %s [%s]. This is a wrong match. Continuing..."
                                         % (
@@ -1927,6 +1939,13 @@ class PostProcessor(object):
                                     logger.fdebug(
                                         "%s[LONE-VOLUME/NO YEAR][MATCH] Only Volume on watchlist matches, no year present in filename. Assuming match based on volume and title."
                                         % module
+                                    )
+                                    datematch = "True"
+
+                                if datematch == "False" and volume_match_settles_year(cs["WatchValues"], lonevol):
+                                    logger.fdebug(
+                                        "%s[MANGA][VOLUME MATCH] Volume %s matched, so the year in the filename (%s) does not decide this file. Assuming match based on volume and title."
+                                        % (module, watchmatch["series_volume"], watchmatch["issue_year"])
                                     )
                                     datematch = "True"
 
@@ -2080,6 +2099,7 @@ class PostProcessor(object):
                                     "Publisher": av["IssuePublisher"],
                                     "Total": int(av["TotalIssues"]),
                                     "Type": av["Type"],
+                                    "IsManga": series_kind.is_manga(av),
                                     "IsArc": True,
                                 },
                             }
@@ -2119,18 +2139,7 @@ class PostProcessor(object):
                                 nm += 1
                             else:
                                 try:
-                                    if (
-                                        any(
-                                            [
-                                                v[i]["WatchValues"]["Type"] == "TPB",
-                                                v[i]["WatchValues"]["Type"] == "GN",
-                                                v[i]["WatchValues"]["Type"] == "HC",
-                                            ]
-                                        )
-                                        and v[i]["WatchValues"]["Total"] > 1
-                                    ) or all(
-                                        [v[i]["WatchValues"]["Type"] == "One-Shot", v[i]["WatchValues"]["Total"] == 1]
-                                    ):
+                                    if volume_identifies_file(v[i]["WatchValues"]):
                                         if arcmatch["series_volume"] is not None:
                                             just_the_digits = re.sub("[^0-9]", "", arcmatch["series_volume"]).strip()
                                         else:
@@ -2149,14 +2158,7 @@ class PostProcessor(object):
                                     temploc = just_the_digits.replace("_", " ")
                                     temploc = re.sub("[\\#']", "", temploc)
                                 else:
-                                    if any(
-                                        [
-                                            v[i]["WatchValues"]["Type"] == "TPB",
-                                            v[i]["WatchValues"]["Type"] == "GN",
-                                            v[i]["WatchValues"]["Type"] == "HC",
-                                            v[i]["WatchValues"]["Type"] == "One-Shot",
-                                        ]
-                                    ):
+                                    if numbered_by_volume(v[i]["WatchValues"]):
                                         temploc = "1"
                                     else:
                                         temploc = None
@@ -2678,6 +2680,7 @@ class PostProcessor(object):
                                             "Total": 0,
                                             "Type": ofl["format"],
                                             "ComicID": ofl["ComicID"],
+                                            "IsManga": series_kind.is_manga(ofl),
                                             "IsArc": False,
                                         },
                                     }
@@ -2696,7 +2699,7 @@ class PostProcessor(object):
                                     continue
                                 else:
                                     try:
-                                        if ofv["WatchValues"]["Type"] is not None and ofv["WatchValues"]["Total"] > 1:
+                                        if volume_identifies_file(ofv["WatchValues"]):
                                             if watchmatch["series_volume"] is not None:
                                                 just_the_digits = re.sub(
                                                     "[^0-9]", "", watchmatch["series_volume"]
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -77,6 +77,27 @@ def log_scan_summary(module, filename, candidate_count, selected_items, annual_c
     )
 
 
+def volume_identifies_file(watch_values):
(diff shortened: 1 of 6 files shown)


comicarr/postprocessor.py
```diff
@@ -77,6 +77,66 @@ def log_scan_summary(module, filename, candidate_count, selected_items, annual_c
...
+
+    This is :func:`numbered_by_volume` qualified by the run length, because a
+    collected edition only numbers *files* by volume when the run actually has
+    volumes to distinguish: TPB/HC/GN spanning more than one entry, or a
+    One-Shot standing alone.
+
+    Manga is not qualified that way. Volume 1 of a one-volume manga is still
+    identified by its volume, so the run length says nothing useful here.
+    """
+    if watch_values.get("IsManga"):
```
frankieramirez, line 199: This makes every manga series locate the file by series_volume digits, and those digits still go into Int_IssueNumber a few dozen lines down. On MangaDex/MAL that column is the chapter, so a v20 file gets attached to chapter 20 and Have ticks up on the wrong row.

rescan already matches volume files via VolumeNumber and falls back to Int_IssueNumber when there are no volume rows (the ComicVine volumes-as-issues case). Could we do the same split here?

comicarr/postprocessor.py
```diff
@@ -1342,16 +1404,7 @@ def Process(self):
...
-                                any(
-                                    [
-                                        cs["WatchValues"]["Type"] == "TPB",
-                                        cs["WatchValues"]["Type"] == "HC",
-                                        cs["WatchValues"]["Type"] == "GN",
-                                    ]
-                                )
-                                and cs["WatchValues"]["Total"] > 1
-                            ) or all([cs["WatchValues"]["Type"] == "One-Shot", cs["WatchValues"]["Total"] == 1]):
+                            if volume_identifies_file(cs["WatchValues"]):
```
frankieramirez: FileChecker sets booktype to manga for a c/ch/chapter token and then defaults a missing volume to v1. Once this arm is true for every manga series, Chainsaw Man c181 (2023).cbz locates as issue 1 instead of 181.

Skip series_volume when watchmatch["booktype"] == "manga" and fall through to justthedigits. Volume-only files keep booktype "issue".

comicarr/postprocessor.py
```diff
@@ -1916,6 +1942,13 @@ def Process(self):
                                     )
                                     datematch = "True"
 
+                                if datematch == "False" and volume_match_settles_year(cs["WatchValues"], lonevol):
```
frankieramirez: lonevol is "filename volume digits equal ComicVersion", and ComicVersion is unset for manga so the scan defaults it to 1. v01 with a year mismatch gets through (that's the live log). v20 still dies on 20 != 1.

Pass that the file was located by volume, not lonevol. Something like volume_identifies_file(...) and watchmatch["series_volume"] is not None.

tests/unit/test_manga_live_paths.py
```diff
@@ -164,3 +164,43 @@ def test_parse_kwargs_auto_passes_folder_bare_numbers():
...
+    """
+    pp_path = Path(__file__).resolve().parents[2] / "comicarr" / "postprocessor.py"
+    tree = ast.parse(pp_path.read_text(encoding="utf-8"))
+
+    called = {
+        node.func.id
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Call) and isinstance(node.func, ast.Name)
+    }
+    assert "volume_identifies_file" in called, "the folder scan no longer consults the volume predicate"
```
frankieramirez, line 462: These AST checks stay green if Process never calls volume_match_settles_year, or keeps passing lonevol. The helper tests feed volume_matched=True themselves, so they wouldn't catch that either.

Could we assert volume_match_settles_year is in `called`, and add a Process (or extracted year-gate) test for v20 + a year mismatch that expects the issue selected?

tests/unit/test_manga_live_paths.py
```diff
@@ -164,3 +164,43 @@ def test_parse_kwargs_auto_passes_folder_bare_numbers():
...
+
+def test_the_volume_numbered_type_list_has_a_single_owner():
+    """The scan tested a hardcoded TPB/GN/HC/One-Shot list in eight places.
+
+    Each copy silently omitted manga, so a manga volume file failed a different
+    check depending on which copy it reached. The list now lives in one tuple.
+    """
+    pp_path = Path(__file__).resolve().parents[2] / "comicarr" / "postprocessor.py"
+    source = pp_path.read_text(encoding="utf-8")
+    assert source.count('"TPB"') == 1, "the volume-numbered type list has been duplicated again"
```
frankieramirez: nit: this misses a duplicate written as 'TPB', and a comment that mentions "TPB" fails the suite. An AST check that `_COLLECTED_TYPES` is assigned once as ("TPB", "HC", "GN") would pin the actual invariant.

frankieramirez (commented):
The parser change for trailing tags looks right. Folder scan still keys year settle off ComicVersion (v02+ fail the year gate) and still looks up Int_IssueNumber with the volume digits (wrong row on MangaDex).

comicarr/postprocessor.py
azuretek: Yes — and this was the worst of the three, because it silently attaches a file to a row it has nothing to do with rather than just failing to match. Fixed in 2c6dc81 by mirroring rescan, as you suggested.

New `manga_volume_rows()` resolves on `VolumeNumber`, and returns `None` — not an empty list — when the series has no volume rows at all. That distinction is what makes the ComicVine case work: `None` means "this lookup doesn't apply, run your own query", while `[]` means "the series has volumes, just not this one". So volumes-as-issues still goes through `Int_IssueNumber`, which is correct there.

```python
if cs["WatchValues"].get("IsManga") and not chapter_named_file(watchmatch):
    issuechk = manga_volume_rows(cs["ComicID"], watchmatch["series_volume"])
if issuechk is None:
    issuechk = db.select_all(...Int_IssueNumber == fcdigit...)
```

Matching goes through `normalize_volume_number`, so `"v01"` in the filename and `"1"` in the ledger agree — same canonicalisation rescan's `int(float(vol))` gets, without the truncation.

Guarded by `chapter_named_file` too, per your next comment — a chapter file must not take the volume lookup on the strength of a defaulted `v1`.

`TestMangaVolumeRows` covers all five paths: the volume match, zero-padding/marker normalisation, the no-volume-rows deferral, an unusable volume (which never issues a query at all), and a real no-match.

comicarr/postprocessor.py
azuretek: Confirmed against the source, and fixed in 2c6dc81 — both halves of your diagnosis are exactly right:

```python
# filechecker.py:479
if manga_chapter is not None:
    booktype = "manga"

# filechecker.py:1185
if all([len(volume_found) == 0, booktype != "issue"]) or ...:
    issue_volume = "v1"
```

So a chapter file is the *only* manga file that gets `booktype == "manga"`, and it's the one that then receives a `v1` it never stated. `volume_identifies_file` returning True for all manga meant it located on that phantom volume.

Added `chapter_named_file(match)`, which just reads the distinction FileChecker already drew, and gated the locating branch on it:

```python
if volume_identifies_file(cs["WatchValues"]) and not chapter_named_file(watchmatch):
```

so a chapter file falls through to `justthedigits` — which is already the chapter, since `final_issue_number = manga_chapter` when `booktype == "manga"`. Volume-only files keep `booktype "issue"` and are untouched, as you said.

It gates the `VolumeNumber` lookup and the year gate as well, so all three places agree on what "located by volume" means rather than each deciding separately — that divergence is what this PR was trying to end in the first place.

comicarr/postprocessor.py
azuretek: Right, and this is the one I'd have shipped believing it worked — v01 passing by coincidence is exactly the shape of bug that survives testing. Fixed in 2c6dc81.

Took your `volume_identifies_file(...) and watchmatch["series_volume"] is not None` and extracted it, since you offered the year-gate extraction in your test comment and it makes the whole thing testable:

```python
def volume_settles_year_for_match(watch_values, match):
    volume_located = (
        volume_identifies_file(watch_values)
        and not chapter_named_file(match)
        and (match or {}).get("series_volume") is not None
    )
    return volume_match_settles_year(watch_values, volume_located)
```

Added `chapter_named_file` to the composition as well, so a chapter file's defaulted `v1` can't wave itself past the year check on evidence it never carried.

I re-ran the new tests against a simulated `lonevol` gate to be sure they actually catch it:

```
FAILED test_v20_with_a_mismatched_year_is_settled_by_the_volume
FAILED test_a_chapter_file_is_not_settled_by_its_defaulted_volume
```

and `test_v01_is_settled_for_the_same_reason_and_not_by_accident` passes under **both** — which is precisely your point about why v01 was never evidence of anything.

tests/unit/test_manga_live_paths.py
azuretek: Fair, and the criticism applies to the tests I wrote as much as to the checks — they asserted the helpers existed rather than that the scan used them correctly. Fixed in 2c6dc81, both parts.

**AST side.** Added the assertion you asked for, plus one that pins the argument, since "calls it" and "calls it with the right thing" are different claims:

```python
assert "volume_settles_year_for_match" in called
assert "volume_match_settles_year" in called
...
assert "lonevol" not in names, (
    "the year gate is keyed off lonevol again -- manga has no ComicVersion, "
    "so that defaults to 1 and only v01 would pass"
)
```

Also added `chapter_named_file` and `manga_volume_rows` to the `called` set, each with a message naming the regression it guards, so a future removal says what broke rather than just "assertion failed".

**Behavioural side.** Took the extracted-year-gate option — driving the full `Process()` folder scan needed enough mocking that the test would mostly assert the mocks. `TestVolumeSettlesYearForMatch` drives the composed gate with the watchmatch shapes FileChecker actually emits: v20 with a mismatched year (settled), v01 (settled, and noted as not evidence either way), a chapter file's defaulted `v1` (**not** settled), no volume at all (not settled), and a TPB (not settled — the manga-only scope holds).

Verified against a simulated `lonevol` gate: the v20 and chapter-file cases both fail, so the tests catch the regression rather than restating the implementation.

tests/unit/test_manga_live_paths.py
azuretek: Good nit — and it was worse than cosmetic, since a plain comment mentioning TPB would have failed the suite for no reason. Replaced with the AST invariant you described, in 2c6dc81:

```python
definitions = [... ast.Assign targeting _COLLECTED_TYPES ...]
assert len(definitions) == 1
assert isinstance(value, ast.Tuple)
assert [element.value for element in value.elts] == ["TPB", "HC", "GN"]
```

Kept a duplicate check too, but moved it to AST `Constant` nodes equal to `"TPB"` — quote style is invisible to the AST, and a comment isn't a Constant at all, so both of your failure modes go away.

Checked both directions rather than assuming:

- adding a comment mentioning `TPB` and `"TPB"` → **passes** (the old test failed)
- adding `_SNEAKY = ('TPB', 'HC')` → **fails** with `where 2 = len(['TPB', 'TPB'])` (the old test passed)

azuretek:
Both points from your summary are fixed in 2c6dc81, with replies inline on each thread:

- **Year settle keyed off ComicVersion (v02+ failing the gate)** — the gate no longer touches `lonevol`. Extracted as `volume_settles_year_for_match()`, which asks whether the file was located by its *own* volume. Verified the new tests fail against a simulated `lonevol` gate, so v20 is genuinely covered rather than assumed.
- **Int_IssueNumber looked up with the volume digits** — `manga_volume_rows()` resolves on `VolumeNumber`, mirroring rescan, and returns `None` (not `[]`) when the series has no volume rows so the ComicVine volumes-as-issues case still uses the issue query.

Also folded in the third finding — a chapter file no longer locates by the `v1` FileChecker defaults onto it — and both test comments: the AST check now asserts the gate is reached *and* is never handed `lonevol`, and the TPB single-owner check is an AST invariant rather than a source-text count.

Added `.changeset/manga-volume-import-match.md` while I was here, since this one is squarely operator-visible.

Full suite green: **2853 passed**. `ruff check` / `ruff format --check` clean.

frankieramirez (approved):

azuretek:
Rebased onto current main (6df49e3); no content changed in the rebase, and the approval above still describes this tree.

One correction worth making, since it predates the rebase: the reply resolving your two summary points cites `2c6dc81`, and that commit was never on this branch. It was the head of a local branch from before an earlier rebase. The same change is here as `878b164`, "fix(manga): file a volume against its volume row, not a same-numbered chapter", and the diff is byte-identical to the one that reply describes. Nothing is missing, but the SHA sent you somewhere that does not exist and I would rather say so than leave it.

Pre-rebase head kept at `azuretek/comicarr@backup/manga-import-20260903`.

Green after the rebase: full unit suite (2878 passed), and `ruff check` plus `ruff format --check` on `comicarr/`.

azuretek:
Rebased onto current main (`f8fc9b4`), which this branch was conflicting with. **The force-push dismissed your approval above** — the tree is the same change, but GitHub reset the review, so it needs re-approving.

Nothing was dropped and nothing was rewritten by hand. The only conflict was in `tests/unit/test_manga_live_paths.py`, where main and this branch each appended a new test at the same point; both are kept.

Old to new:

| before | after |
| --- | --- |
| `878b164` | `338a427` chore: add changeset for the manga volume import fixes |
| `2d20106` | `c6e4d55` fix(manga): file a volume against its volume row, not a same-numbered chapter |
| `235c526` | `1e3ac19` fix(manga): give the folder scan one owner for volume-numbered series |
| `990d142` | `ef6b27d` fix(manga): number a watchlisted manga file by its volume during folder scan |
| `14496c7` | `a932889` fix(manga): parse volume/chapter tokens followed by release tags |

Pre-rebase head kept at `azuretek/comicarr@backup/manga-import-20260905`.

Green after the rebase: 2988 passed in the unit suite, and `ruff check` plus `ruff format --check` clean on `comicarr/`.

PR #821, opened 2026-09-01
fix(ui): name a ledger by what its rows are, not by the content kind

### LINKED ISSUE #817: Series table labels a volume ledger "Chapters" and leaves the type column blank
### Summary

On a manga series the series table labels the ledger **"Chapters"** and leaves the type column blank on every row, even when every row is a **volume**.

A licensed manga catalogued through ComicVine models the English **volumes** as its issues. So a series like One-Punch Man is 33 *volume* rows, and the table renders:

```
Chapters 33          <- wrong noun
(type column empty on all 33 rows)
```

### Why this matters beyond cosmetics

Volumes and chapters are **separate acquisition targets**, and the blended monitor mode searches both. So this is the one place in the UI where the distinction is load-bearing: the table tells an operator the series is tracked by chapter when it is in fact tracked by volume, which is exactly the information needed to judge whether monitoring is set up correctly.

### Reproduce

1. Add a manga series whose issue rows carry `volumeNumber` and no `chapterNumber`.
2. Open the series detail page.
3. The ledger header reads `Chapters N`, and the per-row type column is blank.

### Also observed

The table prints **"grouped by arc"** unconditionally whenever a series has no annuals — including on ledgers where every arc cell is empty.

### Expected

The label and the per-row badge should be derived from what the rows actually contain (`chapterNumber` / `volumeNumber`), not from the content kind:

- rows with volume numbers only -> `Volumes`
- rows with chapter numbers -> `Chapters`
- a ledger holding both -> `Volumes & chapters`, with a count of each
- neither populated (ordinary comics) -> `Issues`, unchanged

### Environment

Reproduced on a ComicVine-sourced manga series with a populated volume ledger.

azuretek:
Fixes #817.

## Problem

The series table named the ledger from the content kind rather than from its rows, so a manga series whose rows are **volumes** rendered as `Chapters N` with an empty type column on every row.

That matters because volumes and chapters are separate acquisition targets and the blended monitor mode searches both — the table was reporting the wrong one of the two things an operator needs to see.

## Change

One commit, frontend only.

Derive the label and the per-row badge from the row's own `chapterNumber` / `volumeNumber`:

- Chapter wins over volume, because a chapter row also carries the volume containing it while a volume row has no chapter number.
- A ledger holding both reads `Volumes & chapters` and prints a count of each.
- Also stops printing "grouped by arc" on a ledger where every arc cell is empty; it previously printed unconditionally whenever a series had no annuals.

**Comic behaviour is unchanged.** With neither number populated the table still reads `Issues` and the type column stays blank except for annuals.

## Test

`frontend/tests/pages/SeriesDetailPage.test.tsx`, 128 lines, covering each label case, the mixed ledger, the per-row badge, the arc-grouping condition, and the unchanged comic rendering.

| | |
|---|---|
| With the change | **24 passed** |
| Source reverted, tests kept | **7 failed, 17 passed** |

The 17 that still pass are the comic-behaviour controls — they assert *preserved* rendering, so they correctly pass on both sides. The 7 failures are the new behaviour.

Backend suite is unaffected and matches the clean-tree baseline exactly (**2827 passed, 8 failed, 1 skipped**; the 8 are `tests/integration/test_schema_migration_dialects.py`, which needs a database URL not available in this environment and fails identically on `main`).

## Base

Branches from `main` and depends on nothing else.

## Summary by CodeRabbit

* **New Features**
  * Added clearer ledger labels for volumes, chapters, annuals, and issues.
  * Added volume and chapter counts when applicable, including mixed ledgers.
  * Updated row badges to show each item’s detected type.
  * Added arc-grouping details only when arc information is available.

* **Bug Fixes**
  * Improved labeling for mixed and incomplete issue data with sensible fallback labels.
  * Correctly handles manga volumes represented as issues while preserving comic ledgers as “Issues.”


azuretek:
Added a changeset here too — `.changeset/manga-ledger-label.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

frontend/src/pages/SeriesDetailPage.tsx
```diff
@@ -392,6 +424,34 @@ export default function SeriesDetailPage() {
...
+  // keeps a synthesised placeholder ledger reading as chapters, as before.
+  const ledgerLabel =
+    volumeCount && chapterCount
+      ? "Volumes & chapters"
+      : volumeCount
+        ? "Volumes"
+        : chapterCount
+          ? "Chapters"
+          : isManga
+            ? "Chapters"
```
frankieramirez, line 455: A ComicVine volume series like One-Punch Man still shows up as Chapters, and the type column is empty. CV keeps the volume in Issue_Number (API `number`) and never writes VolumeNumber, so every row lands here. Treat a manga row as a volume when `number` is set and `chapterNumber`/`volumeNumber` are empty (MangaDex chapters already have `chapterNumber`), and point the CV test at `number` only expecting Volumes.

azuretek: Fixed in eb60aeb.

You were right that every CV row lands in the `return null` — and the type column being blank was the same miss, since the badge is driven off the same lookup. `getLedgerKind` now takes the content kind and treats a bare number as a volume on a manga ledger only; on a comic ledger the same row stays an issue, so `Invincible` is untouched.

```ts
if (isManga && hasLedgerNumber(issue.number ?? issue.Issue_Number)) {
  return "volume";
}
```

Kept it below the `volumeNumber` branch so an explicit VolumeNumber still wins, and below `chapterNumber` so a MangaDex row that carries both is still a chapter.

Pointed the CV test at `number` only, as you asked — it fails without the source change (label reads `Chapters`, so it was passing for the wrong reason before). Added a separate case for rows that do carry `volumeNumber`, since that path was what the old test actually covered and I did not want to lose it.

One thing your comment surfaced that I had to fix as well: `ledgerRow` sets `Issue_Number` on every row, so the "carrying no numbers" fallback test would have started reading as Volumes. That test now genuinely carries no number under either spelling — it was only passing before because the helper's default happened to be ignored.

azuretek:

The replies on this PR cite `eb60aeb`, which was the head before the rebase. That commit is now `0a0b201`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/ui-ledger-name-20260903`.

PR #822, opened 2026-09-01
fix(recovery): close a band row whose work is provably already done

### LINKED ISSUE #818: A parked manual_review row is never cleared after its work provably completes
### Summary

A row parked in `manual_review` / `failed` stays in the attention band **forever**, even after the work it is asking about has demonstrably completed — the issue is `Downloaded`, the file is in the library, and there is nothing left for an operator to do.

### Why nothing ever clears it

Two mechanisms could, and each is structurally excluded:

- `replay_pipeline` builds its work list from `journal.read_open()`. Both `manual_review` and `failed` are **terminal**, so a parked row is invisible to recovery on every future startup.
- `reconcile_existing_excluded_rows` *does* walk exactly these rows — but skips every one whose reason is **actionable**. An actionable reason is precisely what put the row in front of an operator in the first place.

So nothing re-examines whether the question is still worth asking.

### What the operator is offered

Every option is wrong once the work is already done:

| Offered action | Effect when the work is already complete |
|---|---|
| Import | re-runs a move whose source is legitimately gone |
| Search again | re-downloads a file already in the library |
| Stop wanting | discards a volume that is owned |

The row cannot be resolved truthfully, so it accumulates.

### Reproduce

1. Let a download complete and its file be placed into the library.
2. Arrange for the marker write to be lost (any failure after the move commits) so the obligation is parked in `manual_review`.
3. Restart. The issue reads `Downloaded` and the file is present.
4. The band row remains, on this and every subsequent startup.

### Expected

Before leaving a parked row in the band, check the **destination**: when the obligation's own issue is `Downloaded` and its `Location` resolves to an existing file beneath the series root, the work is done — close the row and drop it out of the band.

### Relationship to #810

This is the same rule one stage later. #810 stops a `post_processing` row with positive placement evidence from being re-driven forever; this stops a row **already parked** by that class of failure from being stranded forever. Both are destination probes through the same evidence gate.

azuretek:
Fixes #818.

> **Builds on #814.** The first two commits in this branch are that PR; only the last commit is new here. Once #814 merges, this diff reduces to that single commit.

## Problem

A `manual_review` / `failed` row exists to ask an operator to act, and nothing ever re-examines whether the question is still worth asking. Both stages are terminal, so `replay_pipeline` (which reads `journal.read_open()`) never sees the row; `reconcile_existing_excluded_rows` does walk it but skips every row whose reason is actionable — which is exactly why it was parked.

The result is a row asking for an action no operator can usefully take: the issue is already `Downloaded`, the file is already in the library, and the three offered outcomes all make things worse (re-run a move whose source is gone, re-download a file we own, or discard a volume we own).

## Change

Check the **destination** before leaving the row in the band: when the obligation's own issue is `Downloaded` and its `Location` resolves to an existing file beneath the series root, the work is done, so close the row and drop it out of the band.

This is #814's rule applied one stage later, through the same `has_verified_library_file` gate. Both are **destination** probes on purpose — a surviving *source* is undecidable under copy/hardlink/softlink `FILE_OPTS`, but a verified file in the library proves the move committed in every mode.

### Two implementation notes worth flagging in review

**It stamps the R9 resolution rather than advancing the stage, deliberately.** The lattice is forward-only and `manual_review` (rank 55) sits *above* `post_processed` (50), so `record_transition` would log a no-op and leave the row admitted — the close would appear to work and change nothing. Stamping `imported` and deleting the nzblog anchor reaches the same end state a successful operator "Import" produces, without re-running the import.

**Fails closed.** Any missing id, unreadable row, unresolvable path, or failed stamp leaves the row exactly as it was, still in front of the operator.

Also extracts the nzblog anchor delete into `_delete_nzblog_anchor` so the close paths share one implementation of the story-arc and nzbname disambiguation, rather than a second copy.

## Test

| | |
|---|---|
| Full suite on this branch | **2836 passed, 8 failed, 1 skipped** |
| New tests, source reverted | **3 failed, 8 passed** |

The reverted run is a genuine assertion failure, not a collection error, and the 8 that still pass are controls asserting that a parked row whose issue is **not** `Downloaded` is correctly left alone — so the gate is proven scoped rather than merely permissive.

### DIFF
--- a/comicarr/app/downloads/recovery.py
+++ b/comicarr/app/downloads/recovery.py
@@ -551,7 +551,7 @@ def _obligation_already_fulfilled(row):
     return has_verified_library_file(series.get("ComicLocation"), issue.get("Location"))
 
 
-def finalize_post_processing(row, payload=None):
+def finalize_post_processing(row, payload=None, already_fulfilled=None):
     """Resolve a row already inside post-processing. The `moved` marker is the
     discriminator for whether the destructive move committed — there is NO
     SOURCE probe anywhere on this path (a surviving source is undecidable in
@@ -589,7 +589,9 @@ def finalize_post_processing(row, payload=None):
         return "moved-finish-dbfacts"
 
     if stage == journal.POST_PROCESSING:
-        if _obligation_already_fulfilled(row):
+        if already_fulfilled is None:
+            already_fulfilled = _obligation_already_fulfilled(row)
+        if already_fulfilled:
             _finish_db_facts_only(rkey, row, payload)
             logger.info(
                 "[RECOVERY] %s was `post_processing`, but its issue is already "
@@ -729,16 +731,23 @@ def _resolve_row(snapshot_row, probes=None, pp_cap=None):
     if cur_stage == journal.MOVED:
         return finalize_post_processing(row, payload=payload)
     if cur_stage == journal.POST_PROCESSING:
-        if pp_cap is not None and pp_cap.get("count", 0) >= _MAX_INLINE_PP_REDRIVE_PER_PASS:
-            logger.warn(
-                "[RECOVERY] %s is `post_processing` but the inline PP re-drive "
-                "cap (%d) for this replay pass is reached — DEFERRING; it "
-                "resumes next startup (replay is idempotent/re-runnable)." % (rkey, _MAX_INLINE_PP_REDRIVE_PER_PASS)
-            )
-            return "skip-pp-cap-deferred"
-        if pp_cap is not None:
-            pp_cap["count"] = pp_cap.get("count", 0) + 1
-        return finalize_post_processing(row, payload=payload)
+        # A provably-fulfilled obligation costs only the same DB facts as
+        # `moved`, so — per this cap's own contract — it is NOT charged to the
+        # INLINE re-drive budget. Otherwise a backlog of already-done rows
+        # would need one restart per five to drain.
+        already_fulfilled = _obligation_already_fulfilled(row)
+        if not already_fulfilled:
+            if pp_cap is not None and pp_cap.get("count", 0) >= _MAX_INLINE_PP_REDRIVE_PER_PASS:
+                logger.warn(
+                    "[RECOVERY] %s is `post_processing` but the inline PP re-drive "
+                    "cap (%d) for this replay pass is reached — DEFERRING; it "
+                    "resumes next startup (replay is idempotent/re-runnable)."
+                    % (rkey, _MAX_INLINE_PP_REDRIVE_PER_PASS)
+                )
+                return "skip-pp-cap-deferred"
+            if pp_cap is not None:
+                pp_cap["count"] = pp_cap.get("count", 0) + 1
+        return finalize_post_processing(row, payload=payload, already_fulfilled=already_fulfilled)
 
     if cur_stage == journal.DOWNLOADED:
         item = _pp_item_from_row(row, payload)
--- a/comicarr/app/downloads/recovery.py
+++ b/comicarr/app/downloads/recovery.py
@@ -32,16 +32,18 @@ never aborting the loop.
 """
 
 import time
+import traceback
 
 from sqlalchemy import and_, delete, or_, select
 
 import comicarr
 from comicarr import db, logger
+from comicarr.app.acquisition.evidence import has_verified_library_file
 from comicarr.app.attention import ManualReview, record
 from comicarr.app.downloads import journal, recovery_classify
 from comicarr.app.downloads.ddl_commands import DDLCommand, DDLCommandError
 from comicarr.app.downloads.pp_commands import PostProcessCommandError, validate_postprocess_item
-from comicarr.tables import ddl_info, nzblog, pipeline_journal, snatched, storyarcs
+from comicarr.tables import comics, ddl_info, issues, nzblog, pipeline_journal, snatched, storyarcs
 
 _ENQUEUE_THROTTLE_SECONDS = 0.05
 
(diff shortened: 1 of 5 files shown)


azuretek:

azuretek:
Added a changeset here too — `.changeset/recovery-band-fulfilled.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

comicarr/app/downloads/recovery.py
```diff
@@ -468,22 +470,152 @@ def _resume_item_from_row(row, payload):
...
+    if not _obligation_already_fulfilled(row):
+        return False
+    rkey = row.get("release_key")
+    if not rkey:
+        return False
+    if payload is None:
+        payload = journal.load_payload(row.get("payload_json"))
+    with db.get_engine().begin() as conn:
+        _delete_nzblog_anchor(conn, row, payload)
+        return bool(journal.stamp_resolution(rkey, journal.STATUS_IMPORTED, conn=conn))
```
frankieramirez: SQLAlchemy commits this block when stamp_resolution returns False, so we wipe the nzblog row and leave the journal unresolved. stamp_resolution does that (no raise) if the row is gone or already resolved.

Stamp first on this connection, and only delete the anchor if the stamp returns True.

Add a test that forces stamp_resolution to return False and asserts the nzblog row is still there.

comicarr/app/downloads/recovery.py
```diff
@@ -468,22 +470,152 @@ def _resume_item_from_row(row, payload):
...
+    if not issueid:
+        return False
+    try:
+        issue = db.select_one(
+            select(issues.c.Status, issues.c.Location, issues.c.ComicID).where(issues.c.IssueID == str(issueid))
+        )
+    except Exception as e:
+        logger.warn("[RECOVERY] fulfillment lookup failed for issue %s: %s" % (issueid, e))
+        return False
+    if not issue or issue.get("Status") != "Downloaded":
```
frankieramirez: This gate is issue-wide (Downloaded plus a file under the series root) while nzblog is unique on (IssueID, PROVIDER). Closing a parked row for issue X deletes that provider's only nzblog row, and the same startup then mark_done's a still-open DDL sibling that shares the key (DDL release_keys carry a discriminant).

Close the parked row, and skip the nzblog delete while another open journal row shares the issue and provider.

tests/unit/test_recovery_replay.py
```diff
@@ -1561,3 +1562,178 @@ def test_done_signal_without_placement_absent_probe_enqueues_pp_not_done(queues)
...
+    _placed_issue(tmp_path, "904", "Series v04.cbz", create=False, location=str(stray))
+    _pp_obligation(tmp_path, "904", "Series.v04.cbz")
+    calls = _capture_reimport(monkeypatch)
+
+    recovery.replay_pipeline(probes=_probe("complete"))
+
+    assert calls.get("ran") is True
+
+
+def test_fulfilled_backlog_larger_than_the_cap_drains_in_one_pass(queues, tmp_path, monkeypatch):
```
frankieramirez, line 1811: Every row here is already fulfilled, so pp_cap never increments and skip-pp-cap-deferred never runs. This still catches charging fulfilled rows to the cap. I'd add a mixed case so we also catch a later change that checks the cap before fulfillment: MAX unfulfilled re-drives, then one fulfilled row that still reaches post_processed.

comicarr/app/downloads/recovery.py
azuretek: Confirmed and fixed in 76fd0bf. `stamp_resolution` returns False on three no-raise paths — row absent, stage not in `BAND_STAGES`, status already in `RESOLVED_STATUSES` (journal.py:949-959) — so the `begin()` block exits cleanly and commits the delete regardless.

Stamped first, delete only on success:

```python
with db.get_engine().begin() as conn:
    if not journal.stamp_resolution(rkey, journal.STATUS_IMPORTED, conn=conn):
        return False
    if not _anchor_shared_with_open_row(conn, rkey, row):
        _delete_nzblog_anchor(conn, row, payload)
    return True
```

The `_anchor_shared_with_open_row` call is the other thread; the ordering change is independent of it.

Test added as you asked — `test_failed_stamp_leaves_the_nzblog_anchor_in_place` monkeypatches `stamp_resolution` to False and asserts the nzblog row is still there and `status` is still None. It fails against the old ordering.

Worth naming why this was the bad direction to fail in: the anchor is the only thing a later pass can reconstruct the obligation from, so committing the delete beside an unresolved row leaves it open with nothing to rebuild it — strictly worse than the whole close not happening.

comicarr/app/downloads/recovery.py
azuretek: Fixed in 76fd0bf, as you described — close the parked row, skip the delete while an open sibling shares the key.

```python
def _anchor_shared_with_open_row(conn, rkey, row):
    ...
    sibling = conn.execute(
        select(pipeline_journal.c.release_key)
        .where(pipeline_journal.c.release_key != rkey)
        .where(pipeline_journal.c.issueid == str(issueid))
        .where(pipeline_journal.c.provider == provider)
        .where(pipeline_journal.c.stage.in_(journal.OPEN_STAGES))
        .limit(1)
    ).fetchone()
```

`OPEN_STAGES` rather than "not terminal" so a band sibling — which is parked for an operator, not waiting on this anchor — does not pin it either. The anchor then goes with whichever obligation finishes last.

It fails toward keeping the anchor: a lookup error returns True and logs, because a surviving anchor is recoverable on the next pass and a deleted one is not.

Two tests: an open sibling on the same (issueid, provider) asserts the anchor survives *and* that the parked row still reaches `imported` — the close is not what I wanted to give up. A `post_processed` sibling is the control and must not hold the anchor. The open-sibling one fails without the guard.

tests/unit/test_recovery_replay.py
azuretek: You're right — I checked, and `pp_cap["count"]` never increments in that test, so the deferral branch is unreachable and the cap is never hit.

Added `test_fulfilled_row_still_closes_after_the_cap_is_exhausted` in 76fd0bf, the mixed shape you described: `_MAX_INLINE_PP_REDRIVE_PER_PASS` unfulfilled rows spend the budget in full, then one fulfilled row that must still reach `post_processed`.

Ordering it deterministically needed one extra step — `_insert_journal` hardcodes `updated_date`, and `read_open()` orders oldest-first, so every row ties. The fulfilled row gets its date bumped a day so it is read last, after the budget is gone.

```python
assert len(redriven) == _MAX_INLINE_PP_REDRIVE_PER_PASS, "the cap must still bound real re-drives"
assert fulfilled not in redriven, "a fulfilled obligation must never be re-imported"
assert summary["actions"].get("skip-pp-cap-deferred") is None
assert _journal_row(fulfilled)["stage"] == journal.POST_PROCESSED
```

Verified by moving `_obligation_already_fulfilled` below the cap check: the new test fails, and so does the existing backlog one — so the pair covers the ordering from both sides.

azuretek:

The replies on this PR cite `76fd0bf`, which was the head before the rebase. That commit is now `46c4561`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/recovery-band-20260903`.

azuretek:
Rebased onto current main (`f8fc9b4`). #849 changed what this PR should contain, so this is not a straight replay — here is exactly what moved and why.

**Two of the six commits are now superseded by #849 and have been dropped:**

- `1d10cbd` fix(recovery): close a post_processing obligation that is provably fulfilled
- `43008a5` fix(recovery): do not charge a fulfilled obligation to the PP re-drive cap

#849 landed both. `_postprocess_completion.obligation_already_fulfilled()` is the same destination-evidence gate this PR proposed, and `postprocessing.recover()` checks it before the budget test, so a fulfilled row already finishes on DB facts alone and is already not charged to `_MAX_INLINE_PP_REDRIVE_PER_PASS`. Re-filing either would be a duplicate.

**What #849 did not take, and what this PR is now only about:**

Closing a *terminal* band row. `recover()` returns `ignored` for anything not in `{MOVED, POST_PROCESSING}`, and `replay_pipeline` builds its work list from open stages, so a `manual_review` / `failed` row whose issue is already `Downloaded` with a verified library file is still never revisited. It sits in Needs Attention forever, and all three offered actions make things worse. That is the case this closes.

Retargeted onto the new structure rather than kept as-was:

- `close_fulfilled_band_row()` now calls `_postprocess_completion.obligation_already_fulfilled()` instead of carrying its own copy of that gate.
- The nzblog anchor predicate is factored out of `finish_obligation()` into `_postprocess_completion.delete_anchor(conn, row, payload)`, so the band close and the obligation close share one predicate rather than restating it. `finish_obligation()` is otherwise unchanged.
- `_anchor_shared_with_open_row()` and the stamp-before-delete ordering are unchanged in intent; both are folded into the single `close_fulfilled_band_row()` commit now that the helper it depended on lives elsewhere.

Old to new, for the three that remain:

| before | after |
| --- | --- |
| `46c4561` | `d7cfb23` fix(recovery): stamp before deleting, and keep an anchor a sibling needs |
| `8392918` | `f93faa1` chore: add changeset for closing a fulfilled attention-band row |
| `2cadaf8` | `7ecd37f` fix(recovery): close a band row whose work is provably already done |

`756e61f` (style: satisfy ruff check and ruff format) is also gone; it only existed to format code that no longer exists in that form. Lint is clean without it.

Pre-rebase head kept at `azuretek/comicarr@backup/recovery-band-20260905`.

Green after the rebase: 2955 passed in the unit suite, `ruff check` and `ruff format --check` clean on `comicarr/`. I also checked the three tests here fail without the fix rather than only asserting its shape: neutering `close_fulfilled_band_row` fails `test_reconcile_closes_parked_row_whose_work_is_already_done`, `test_open_sibling_on_the_same_anchor_keeps_it` and `test_terminal_sibling_does_not_keep_the_anchor`; disabling only the shared-anchor guard fails `test_open_sibling_on_the_same_anchor_keeps_it` alone.

PR #823, opened 2026-09-01
fix(manga): tag manga volumes on import, as volumes rather than issues

### LINKED ISSUE #819: Manga volumes import untagged, and are tagged as issues when they are tagged
### Summary

Manga volumes come out of the import path **untagged**, and when they are tagged they are tagged as *chapters of a volume named after the series*. Three separate defects stack here, plus one general tagging bug that is not manga-specific.

### 1. `_process_manga` never invokes the metatagger

`_process_manga` places a downloaded file and marks its ledger row, but never calls the tagger. A manga volume therefore lands in the library with **no `ComicInfo.xml`** while every comic beside it has one. The only remedy is a manual bulk re-tag afterwards.

### 2. When tagged, a volume is written as an issue

A licensed manga is catalogued by ComicVine as a Series whose "issues" are the English **volumes**. Meta-tagging wrote that shape through unchanged:

```xml
<Number>1</Number>
<Volume>2014</Volume>
```

A reader consuming `ComicInfo` then files every volume as a chapter of one volume named after the series, and the volume structure the files were actually published in is lost.

### 3. `-m "issue="` does not survive the online overlay

The obvious fix — asking ComicTagger to clear the issue number — silently does not work, and the reason is a good diagnostic:

- `create_local_metadata()` ends with `md.overlay(opts.metadata)` under the comment *"finally, use explicit stuff"*.
- `process_file_cli()` **then** runs `md.overlay(cv_md)`.
- ComicVine catalogues an English manga volume as an issue, and `mapCVDataToMetadata` sets `metadata.issue` unconditionally.

So `v7` comes back numbered `7` and overwrites the clear. The asymmetry that makes this diagnosable: the *same* `-m` string sets `volume=N` and `issue=`, and **volume survives while issue does not** — because `metadata.volume` is only set when `use_series_start_as_volume` is on.

Every volume therefore lands tagged `<Number>7</Number>`. Recovering from it costs a full delete-and-rebuild of the series, which throws away read progress.

### 4. Not manga-specific — tagging reduces a file's permissions

ComicTagger cannot write into an existing archive, so `run()` copies the file into a temp dir and ComicTagger emits a **new** file (converting `.cbr` -> `.cbz` on the way). That file is created with tempfile semantics — **`0600`** — rather than inheriting the library file's mode, and is then moved back into the library.

A tagged issue therefore ends up **less readable than every untagged issue beside it**, with nothing logged to say so. This affects any tagged file, not just manga.

### Reproduce

1. Add a manga series with a populated volume ledger and download a volume.
2. Observe the placed file has no `ComicInfo.xml`.
3. Tag it manually: the tags read `<Number>N</Number>` with `<Volume>` set to the series year.
4. Compare the tagged file's mode against an untagged file beside it — `0600` vs the library's mode.

### Expected

- The manga import path tags, as the comic path does.
- A manga volume is tagged as **volume N with no issue number** — the file *is* the book.
- The issue clear is applied where it survives the online overlay.
- Tagging never leaves a file less readable than it was.

azuretek:
Fixes #819.

> **Builds on #816.** The first three commits in this branch are that PR; the last four are new here. Once #816 merges, this diff reduces to those four.

## Problem

Manga volumes were imported untagged, and when tagged were written as issues of a volume named after the series. See the linked issue for the full chain; the short version is four independent defects, one of which is not manga-specific.

## Change — four commits

**1. `fix(manga): tag volumes as volumes, not chapters of the series year`**

A manga volume is a book: the file *is* volume N and carries no issue number. Resolve that from the ledger and write it, clearing the issue number with an empty override — the same mechanism the existing `cvers = "volume="` default already relies on.

Both facts are read from their existing owners rather than restated: `series_kind.is_manga` decides the Series, `ledger.normalize_volume_number` decides the number. Resolution is best-effort, so a missing row, a ledger with no volume numbers, or a non-manga Series all leave the periodical tag shape **byte-identical**.

**2. `fix(metatag): keep a tagged file as readable as it was before tagging`** — *not manga-specific; cherry-pickable on its own*

Capture the mode before tagging starts (the last point it can be read, since `FILE_OPTS = move` deletes the `.cbr` once it becomes a `.cbz`) and apply it to the tagged file before `run()` returns.

`ENFORCE_PERMS` already owns file modes via `filechecker.setperms()`, so when it is on this **defers to it** exactly as the six other placement sites do, rather than becoming a second opinion on `CHMOD_FILE`. When it is off, the user asked for modes to be left alone — which means left alone, not quietly reduced.

**3. `fix(manga): clear the issue number after the online overlay, not before it`**

Re-apply the clear once tagging is done: a second save pass with no `-o` reads back the tags just written, overlays the empty issue, and saves. `overlay()` maps `""` to `None` and the ComicInfo writer skips a `None` field, so `<Number>` is **removed** rather than emptied. No other field is passed and no online fetch is repeated.

**Scoped to the manga branch on purpose.** Comicarr also sends a bare `volume=` whenever `comversion` is `None`, so re-applying `opts.metadata` after the overlay *in general* would strip the start-year `<Volume>` a periodical comic legitimately carries.

**4. `fix(manga): tag a manga file on import, as the comic path does`**

The comic path tags at the download location and then places whatever ComicTagger returns, because tagging can rename the file. The manga path now follows the same order, which required resolving the ledger row before placement — the tagger needs the `IssueID` to know which volume the file is, and it must run while the file is still at its download location.

That lookup is extracted as `_match_manga_issue()`; it only reads `parsed` and the series' issue rows, so moving it earlier **changes nothing about which row is chosen**. Tagging is gated on the same condition the comic path uses (`ENABLE_META` or `CBR2CBZ_ONLY`).

## Safety

Every one of the four is **best effort**: an unresolvable row, absent configuration, a missing `comictaggerlib`, an unreadable original mode, a failing chmod, or any tagger sentinel places the file *untagged* rather than costing the import. Nothing here can fail an import that previously succeeded.

## Test

| | |
|---|---|
| Full suite on this branch | **2896 passed, 8 failed, 1 skipped** |
| New tests, source reverted | **32 failed, 53 passed** |

The reverted run gives genuine assertion failures, and the 53 that still pass are controls — including the ones asserting the periodical tag shape is unchanged and that a failing subprocess never breaks tagging.


azuretek:

azuretek:
Added a changeset here too — `.changeset/manga-volume-tagging.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

frankieramirez (commented):
A few issues on the live manga row types. ComicVine volumes never enter the volume-book branch, and MangaDex ids get sent to ComicVine --id so tagging aborts.

comicarr/cmtag.py
```diff
@@ -7,7 +7,164 @@
...
+
+    from comicarr import db, series_kind
+    from comicarr.app.manga import ledger
+    from comicarr.tables import comics, issues
+
+    try:
+        row = db.select_one(select(issues.c.ComicID, issues.c.VolumeNumber).where(issues.c.IssueID == issueid))
+        if row is None or "VolumeNumber" not in set(row.keys()):
+            return None
+        volume = ledger.normalize_volume_number(row["VolumeNumber"])
```
frankieramirez: MangaDex chapter rows store the containing book in VolumeNumber, so Chainsaw Man 165 gets tagged as volume 18 with Number stripped. ComicVine licensed volumes leave VolumeNumber null, which means One-Punch Man v7 never hits this branch and keeps Number 7 of the series year. After is_manga, return None when ChapterNumber is set, otherwise use VolumeNumber or Issue_Number.

comicarr/cmtag.py
```diff
@@ -7,7 +7,164 @@
...
+            comicarr.CONFIG.CT_SETTINGSPATH,
+            "--type",
+            style,
+            "-m",
+            "issue=",
+            filepath,
+        ]
+        try:
+            p = subprocess.Popen(script_cmd, stdout=subprocess.PIPE, text=True, stderr=subprocess.STDOUT)
+            out, err = p.communicate()
```
frankieramirez: If ComicTagger wedges here the post-processor never comes back and later files in the job never place. Give communicate() a timeout, kill the child, and return False.

comicarr/postprocessor.py
```diff
@@ -4020,6 +4037,129 @@ def nzb_or_oneoff_pp(self, tinfo=None, manual=None):
...
+            return filepath
+        if not (getattr(config, "ENABLE_META", False) or getattr(config, "CBR2CBZ_ONLY", False)):
+            return filepath
+
+        try:
+            from . import cmtag
+
+            pcheck = cmtag.run(
+                self.nzb_folder,
+                issueid=issueid,
```
frankieramirez, line 4275: MangaDex IssueIDs look like md-xxx-ch165. cmtag.run always sends -o --id, ComicVine 404s, and we place the file untagged. Skip -o/--id unless the id is a ComicVine number.

tests/unit/test_manga_postprocessor.py
```diff
@@ -601,3 +610,520 @@ def test_unmatched_file_logs_warning(self, tmp_path):
...
+    @staticmethod
+    def _run_source():
+        import inspect
+
+        from comicarr import cmtag
+
+        return inspect.getsource(cmtag.run)
+
+    def test_run_resolves_a_manga_volume(self):
+        assert "manga_volume_for_issue(issueid)" in self._run_source(), (
```
frankieramirez, line 1379: This stays green if the call is commented out. Import tests also patch cmtag.run, so CI never sees ComicTagger abort on a md- id or write the wrong VolumeNumber. Stub CONFIG/subprocess and assert the -m volume= line plus no -o/--id for md- ids.

comicarr/cmtag.py
```diff
@@ -7,7 +7,164 @@
...
+
+    ENFORCE_PERMS is the configured owner of file modes, so when it is on this
+    defers to filechecker.setperms() exactly as the six other placement sites
+    do. When it is off the user has asked for modes to be left alone -- which
+    means left alone, not quietly reduced -- so the pre-tag mode is restored.
+
+    Returns True when the mode was applied, False when there was nothing to do.
+    """
+    if comicarr.CONFIG.ENFORCE_PERMS:
+        filechecker.setperms(path)
```
frankieramirez: setperms only catches OSError. A bad CHMOD_FILE (config accepts '888') raises ValueError out of run() after the tag already succeeded, and the comic path only catches ImportError, so post-processing dies. Wrap this like the chmod branch, and don't return True when setperms returns False.

comicarr/postprocessor.py
```diff
@@ -1342,16 +1404,7 @@ def Process(self):
...
-                                any(
-                                    [
-                                        cs["WatchValues"]["Type"] == "TPB",
-                                        cs["WatchValues"]["Type"] == "HC",
-                                        cs["WatchValues"]["Type"] == "GN",
-                                    ]
-                                )
-                                and cs["WatchValues"]["Total"] > 1
-                            ) or all([cs["WatchValues"]["Type"] == "One-Shot", cs["WatchValues"]["Total"] == 1]):
+                            if volume_identifies_file(cs["WatchValues"]):
```
frankieramirez: Manual Run goes through the watchlist scan. series_volume digits then look up Int_IssueNumber, which on MangaDex is the chapter number, so Series v20.cbz marks chapter 20 Downloaded. Resolve IsManga files by ChapterNumber then VolumeNumber, same as _match_manga_issue.

comicarr/postprocessor.py
```diff
@@ -4112,6 +4252,24 @@ def _process_manga(self):
...
+
+            if matching is not None:
+                # Best effort: a row without an IssueID cannot be tagged, but it
+                # can still be placed and marked, so this must not raise.
+                try:
+                    tag_issueid = matching["IssueID"]
+                except (KeyError, TypeError, IndexError):
+                    tag_issueid = None
+                if tag_issueid is not None:
+                    filepath = self._metatag_manga_file(filepath, tag_issueid)
```
frankieramirez, line 4426: cmtag.run copies into CACHE_DIR and returns the cache path. place() then moves that copy, so FILE_OPTS=move leaves the download original for the next pass. Comic Process_next tidyups the pre-tag basename after place; do that here too, and skip it if tagging failed and filepath is still the original.

comicarr/postprocessor.py
```diff
@@ -1916,6 +1942,13 @@ def Process(self):
                                     )
                                     datematch = "True"
 
+                                if datematch == "False" and volume_match_settles_year(cs["WatchValues"], lonevol):
```
frankieramirez: lonevol is true when the filename volume equals ComicVersion, usually v1. One-Punch Man v20 still dies on the 2014/2015 year mismatch this was meant to allow, and v1 of a similarly named series can get forced through. Pass whether the book volume actually matched a row.

comicarr/postprocessor.py
```diff
@@ -4020,6 +4037,129 @@ def nzb_or_oneoff_pp(self, tinfo=None, manual=None):
...
+                        select(issues).where(
+                            and_(
+                                issues.c.ComicID == self.comicid,
+                                issues.c.Issue_Number == str(ch_num),
+                            )
+                        )
+                    )
+
+        if matching is None and parsed.get("volume_number") is not None:
+            vol_str = str(parsed["volume_number"])
```
frankieramirez, line 4211: Parser turns v01 into int 1, so this compares VolumeNumber == '1'. MangaDex stores str(chapter.volume), which can be '01', and the file never matches so tagging never runs. Compare with ledger.normalize_volume_number on both sides.

comicarr/cmtag.py
```diff
@@ -7,7 +7,164 @@
...
+        ]
+        try:
+            p = subprocess.Popen(script_cmd, stdout=subprocess.PIPE, text=True, stderr=subprocess.STDOUT)
+            out, err = p.communicate()
+        except Exception as e:
+            logger.warn("%s [MANGA] Unable to clear the issue number on %s [%s]" % (module, filepath, e))
+            return False
+        if "Save complete" not in out:
+            logger.warn("%s [MANGA] Could not clear the issue number on %s : %s" % (module, filepath, out))
+            return False
```
frankieramirez: If CR saves and CBL fails, we return here with Number gone from one tag block and still present in the other, then run() places the file anyway. Keep going through remaining styles and log a partial clear.

comicarr/postprocessor.py
```diff
@@ -2066,6 +2099,7 @@ def Process(self):
                                     "Publisher": av["IssuePublisher"],
                                     "Total": int(av["TotalIssues"]),
                                     "Type": av["Type"],
+                                    "IsManga": series_kind.is_manga(av),
```
frankieramirez, line 2190: Arc rows now carry IsManga and already use volume_identifies_file, but the year-settle on the watchlist path was never copied onto the arc year block. Same 2014/2015 mismatch still rejects manga that arrives as a story arc.

.changeset/manga-volume-tagging.md
```diff
@@ -0,0 +1,5 @@
+---
+"comicarr": patch
+---
+
+Manga volumes are now tagged as volumes rather than as issues of a series-named volume. An imported volume either carried no metadata at all or was written as issue N of a volume named after the series, which is not what a manga volume is — the file *is* volume N and has no issue number. The volume is now read from the ledger and written as the volume, with the issue number cleared. Periodical tagging is untouched: a non-manga series, a missing ledger row, or a ledger with no volume numbers all produce the same tag as before. Tagging also no longer leaves a file less readable than it found it — the pre-tagging permissions are restored afterwards.
```
frankieramirez: nit: this tip vs main also ships the folder-scan match and trailing-tag parse fixes, and #816's changeset isn't an ancestor here. Operators won't see those in What's New unless this summary names them.

comicarr/cmtag.py
azuretek: Fixed in accd3b2, in the order you gave. The resolver was reading the wrong column at both ends.

```python
if ledger.normalize_volume_number(row.get("ChapterNumber")) is not None:
    return None
volume = ledger.normalize_volume_number(row.get("VolumeNumber"))
if volume is None:
    volume = ledger.normalize_volume_number(row.get("Issue_Number"))
```

Chapter first, because a chapter is not a book — and MangaDex chapter rows carrying their containing volume is exactly what made that check load-bearing rather than defensive. Then VolumeNumber, then Issue_Number for the ComicVine shape.

The ComicVine half is the same thing #821 hit in the UI: a licensed manga's English volumes are catalogued as the series' issues, so the volume is the issue number and VolumeNumber is never written. Two different symptoms, one modelling fact.

Tests use a helper carrying every column the resolver selects, plus the CV case and both chapter cases. Reverting to VolumeNumber-only fails the CV and chapter tests.

comicarr/cmtag.py
azuretek: Fixed in accd3b2 — bounded, killed, reaped:

```python
try:
    out, err = p.communicate(timeout=_CLEAR_ISSUE_TIMEOUT)
except subprocess.TimeoutExpired:
    p.kill()
    p.communicate()   # reap, so the killed child is not left a zombie
```

300s, on the reasoning that this is a second save pass over an archive whose tags were just written — no online fetch, no conversion — so it is generous for a large archive on slow storage and still far short of "forever". Happy to change the number if you would rather it were shorter.

Your framing is the part I had wrong: this is a *cosmetic* tag fix, and a cosmetic fix must not be able to stall the pipeline. It goes on to the next style rather than returning, per the other thread.

The test needed care to be a real control — a fake that raises `TimeoutExpired` passes whether or not a timeout was actually requested, since an unbounded wait cannot time out, it just never returns. So it records the timeout argument and asserts `timeouts[0] == _CLEAR_ISSUE_TIMEOUT`. That version does fail when the bound is removed; the first one I wrote did not.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2. Split into a function so the decision can be asserted directly rather than by grepping run()'s source:

```python
def online_tag_options(issueid, module=""):
    if issueid is None:
        return ["-f", "-o"]
    if str(issueid).strip().isdigit():
        return ["-o", "--id", issueid]
    logger.fdebug("... is not a ComicVine issue id -- tagging from local metadata, no online fetch.")
    return []
```

Deliberately not falling back to `-f -o`: that is an online search *by filename*, and for a manga chapter it finds the wrong series more often than the right one. An accurate local tag beats a confident wrong one.

Covered for `md-…-ch165`, `md-…-v07` and `mal-…`; making the numeric check unconditional fails all three.

tests/unit/test_manga_postprocessor.py
azuretek: You're right on both counts, and the second is the sharper one — with `cmtag.run` patched in the import tests, nothing in CI ever saw the argv.

Rather than stub CONFIG and subprocess around the whole of `run()`, I split the two decisions out as small pure functions — `online_tag_options` and `volume_metadata_field` — and assert on their values. Same coverage, and it removes the reason the source assertions existed.

`volume_metadata_field` pins the `-m volume=` line, including that a book's own volume beats `SETDEFAULTVOLUME` forcing 1 and beats a series volume label, with a periodical control that the comic behaviour is untouched. `online_tag_options` pins no `-o`/`--id` for `md-` ids.

I did first write a `-m` test that asserted `"volume=%s" % volume == "volume=7"`, which is the same failure you are describing here in a new place. Deleted it.

comicarr/cmtag.py
azuretek: Confirmed and fixed in accd3b2. `setperms` catches only `OSError` (filechecker.py:2232) while its own `int(comicarr.CONFIG.CHMOD_FILE, 8)` raises `ValueError` on '888'.

```python
try:
    applied = filechecker.setperms(path)
except Exception as e:
    logger.warn("%s Unable to enforce permissions on %s [%s]" % (module, path, e))
    return False
return bool(applied)
```

Both parts: wrapped like the chmod branch, and no longer answering True when setperms returned False.

The bad shape of this one is the timing — it fires *after* tagging has already succeeded, so a cosmetic step was destroying a completed import and everything queued behind it.

Two tests: a `ValueError`-raising setperms returns False rather than propagating, and a setperms returning False is reported rather than claimed. Restoring the bare call fails both.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2, file-level as you asked. `volume_identifies_file` now takes the parsed file, and `matchIT` carries `manga_chapter`/`manga_volume` through — it produced them but dropped them from the match result, so there was nothing else to decide on.

`series_volume` cannot stand in for them for exactly the reason you gave: its `v1` default makes a chapter file and a genuine volume 1 indistinguishable, which is how `c181` became chapter 1 and `v33` became chapter 33.

The rule is stated as "a volume number and no chapter number", matching `is_volume_target`. This branch predates the ledger refactor that gives that rule one owner, so it is spelled out here with a comment pointing at it; when this and the volume-match branch integrate it should call the ledger function instead of keeping a second copy.

All three scan sites (watchlist, arc, one-off) had a matchIT result in hand already, so all three now pass it.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2. Kept the pre-tag path and removed it after a successful place:

```python
if (
    filepath != pre_tag_path
    and getattr(comicarr.CONFIG, "FILE_OPTS", None) == "move"
    and os.path.isfile(pre_tag_path)
):
    os.remove(pre_tag_path)
```

Guarded on `filepath != pre_tag_path` exactly as you said — when tagging failed, `filepath` is still the original and `place()` has already dealt with it.

Did it per file rather than calling `tidyup()`: that works on the whole folder, and this loop places many files from one folder, so it would take files this run has not placed yet. Same intent, narrower blast radius.

A failure to remove warns rather than raising — the file is already correctly in the library at that point, and the cost is one re-process, not a lost import.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2, and this is the same change as on the volume-match branch, where you raised it too — so whichever lands first carries it.

```python
book_volume_matched = volume_identifies_file(watch_values, watchmatch)
...
if datematch == "False" and volume_match_settles_year(cs["WatchValues"], book_volume_matched):
```

Computed where `lonevol` is reset, inside `if datematch == "True"` — so reaching it means the number matched an issue row, and `volume_identifies_file` being true means that number came from the file's own volume token. That is the book-volume match, with ComicVersion out of the path entirely.

Both halves of what you flagged: v20 now settles (a parametrised test covers v01/v06/v20/v100 with `ComicVersion: None`), and a v1 of a similarly named series can no longer be forced through, because the signal is the file's volume matching a row rather than its volume equalling 1.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2. The exact equality is kept as the fast path and falls back to a normalised comparison:

```python
if matching is None:
    for row in db.select_all(...VolumeNumber.isnot(None)...):
        wanted = normalize_volume_number(vol_str)
        if wanted is not None and normalize_volume_number(row.get("VolumeNumber")) == wanted:
            matching = row
            break
```

Both sides through the ledger, as you said — the padding convention has one owner and this is not the place to add a second.

It needs a scan because normalisation cannot be expressed in the `WHERE`, so it is bounded to one series and only runs when the exact match already missed. If that is too loose for a long-running series I can store a normalised column instead, but that is a migration and felt out of scope here.

(On the sibling branch this is `volume_numbers_match`; this branch predates that helper, so it is spelled out with a pointer to it.)

comicarr/cmtag.py
azuretek: Fixed in accd3b2 — every style is attempted, and a partial clear is reported rather than silently passed:

```python
if failed:
    logger.warn(
        "%s [MANGA] Issue number cleared from %s but not %s on %s -- the tag blocks now disagree"
        % (module, cleared or "no", failed, filepath)
    )
    return False
```

The disagreement is the thing worth naming: `run()` places the file either way, so the early return did not prevent a bad outcome, it produced a *worse* one — two tag blocks that disagree about what the file is, which is harder to notice and harder to repair than either block being wrong on its own.

Three tests: both styles clearing, CR failing then CBL still attempted, and CBL failing after CR. Restoring the early return fails the "first style failure" one.

comicarr/postprocessor.py
azuretek: Fixed in accd3b2. You're right that the arc path had the inputs and never got the settle — it was added to the watchlist path and not copied, so the same manga was accepted on the watchlist and rejected when it arrived as part of a story arc.

```python
arc_book_volume_matched = volume_identifies_file(arc_values, arcmatch)
...
if datematch == "False" and volume_match_settles_year(arc_values, arc_book_volume_matched):
```

Placed after the arc's `[LONE-VOLUME/NO YEAR]` block, mirroring the watchlist ordering, so the two paths now read the same way down the whole year sequence.

This one is worth a note in whatever we do about the duplication: `Process()` now carries two near-identical volume/year sequences, and this thread exists because a fix landed in one and not the other. That is a refactor rather than a review fix, so I have left it — say the word and I will open it separately.

.changeset/manga-volume-tagging.md
azuretek: Rewritten in accd3b2. You're right that operators would have seen a tagging note and none of the matching or parsing fixes.

The summary now covers three groups: the tagging shape (including the ComicVine/MangaDex column split and no longer sending a MangaDex id to ComicVine); the folder scan — volume-versus-chapter decided per file, the year-settle on both the watchlist and arc paths, and the `Series Vol.33` title parse; and what is deliberately unchanged, plus the permissions failure no longer aborting post-processing.

Written from what this tip actually ships against main rather than from what this branch's commits say, since that is the difference you are pointing at. Left it as one patch changeset rather than splitting it — What's New reads better as one manga entry than as four — but happy to split if you would rather each fix were separately visible.

azuretek:

The replies on this PR cite `accd3b2`, which was the head before the rebase. That commit is now `8688455`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/manga-tagging-20260903`.

azuretek:
Rebased onto current main (`f8fc9b4`), which this branch was conflicting with. No content changed in the rebase. The only conflict was in `tests/unit/test_manga_live_paths.py`, where main and this branch each appended a new test at the same point; both are kept.

Old to new:

| before | after |
| --- | --- |
| `8688455` | `021f319` fix(manga): find the volume however the series is catalogued |
| `051a32d` | `17885e6` chore: add changeset for manga volume tagging |
| `f4a1da6` | `559ae3d` style: satisfy ruff check and ruff format on this branch |
| `fbdeba1` | `b30acb6` fix(manga): tag a manga file on import, as the comic path does |
| `454d0ea` | `2e028c5` fix(manga): clear the issue number after the online overlay, not before it |
| `a76c267` | `2db783c` fix(metatag): keep a tagged file as readable as it was before tagging |
| `968f999` | `95e056d` fix(manga): tag volumes as volumes, not chapters of the series year |
| `e74a872` | `75c0c28` fix(manga): give the folder scan one owner for volume-numbered series |
| `9751ee0` | `4dab4cd` fix(manga): number a watchlisted manga file by its volume during folder scan |
| `f46d9fd` | `e31dacd` fix(manga): parse volume/chapter tokens followed by release tags |

Pre-rebase head kept at `azuretek/comicarr@backup/manga-tagging-20260905`.

Green after the rebase: 3039 passed in the unit suite, and `ruff check` plus `ruff format --check` clean on `comicarr/`.

azuretek:
Rebased onto current main (`c371ef5`), which this branch was conflicting with.

**#816 merging changed what this PR should contain.** Four of the thirteen commits are now on main through it and have been dropped, and one more has been reduced, so the diff is 5 files instead of 8.

The reduced one is `fix(manga): find the volume however the series is catalogued`. It carried a per-file volume-versus-chapter decision, as a second argument to `volume_identifies_file()`, plus the year-settle that reads it. #816 fixes both by a different route: `volume_identifies_file()` stays series-level and each call site is guarded with `and not chapter_named_file(match)`. That is merged, so this branch now defers to it and no longer touches that function or its tests. Its commit message has been rewritten to say so rather than describing the version that was superseded.

One gap is left behind, and it is the only postprocessor scan change still here: #816 added the year-settle to the watchlist path only, so a manga whose 2014-labelled file matches a 2015 ledger year is accepted on the watchlist and still rejected when the same file arrives as part of a story arc. The arc path now calls main's `volume_settles_year_for_match(arc_values, arcmatch)`. Happy to split that into its own PR if you would rather review it apart from the tagging.

What remains here is the tagging work the title describes: `cmtag.py`, the metatag call on the manga import path, and their tests.

Old to new:

| before | after | |
|---|---|---|
| `535e83e` | dropped | parse volume/chapter tokens followed by release tags |
| `185d808` | dropped | number a watchlisted manga file by its volume during folder scan |
| `67f2250` | dropped | give the folder scan one owner for volume-numbered series |
| `9b4235b` | dropped | keep the folder series name for a tagged chapter-only file |
| `0df6fb5` | `bb6ff10` | tag volumes as volumes, not chapters of the series year |
| `067a913` | `b42976e` | keep a tagged file as readable as it was before tagging |
| `45e01fd` | `e2b1e5e` | clear the issue number after the online overlay, not before it |
| `9e6c5fd` | `fa54e4e` | tag a manga file on import, as the comic path does |
| `fe87dee` | `b089659` | satisfy ruff check and ruff format on this branch |
| `9817501` | `73227a6` | add changeset for manga volume tagging |
| `4b7d52a` | `e0cd0c2` | find the volume however the series is catalogued (reduced, message rewritten) |
| `2079827` | `4d42b8f` | house punctuation in the text this branch adds |
| `42392bc` | `974c12f` | keep a tidy-up from costing an already-successful import |

The four dropped ones were confirmed against main by symbol and by test name, not by subject line: every test function each of them added is present on `origin/main`. The pre-rebase head is kept at `azuretek/comicarr@backup/manga-tagging-20260906`.

Green after the rebase, on the ruff version CI pins (0.15.7):

- `pytest tests/unit -m "not slow"`: 3064 passed, 0 failed. Main's baseline on the same command is 3006, so this adds 58 and removes none.
- `pytest tests/integration -m "not slow"`: 14 passed.
- `ruff check comicarr/`, `ruff format --check comicarr/`, `python scripts/run_ruff.py modern`: all clean.

PR #824, opened 2026-09-01
fix(manga): stop the store-date gate and a title parse from discarding correct volume releases

### LINKED ISSUE #820: Correct manga volume releases are found, then discarded by the store-date gate and a title parse
### Summary

Two independent defects each cause a **correct** manga volume release to be found and then discarded, so volumes read as "not available" when they are indexed and downloadable.

---

## 1. The store-date gate rejects a correct volume

A manga volume search found the right file and threw it away:

```
checking search result: One-Punch Man v06 (2015) (Digital) (BlackManta-Empire)
[CONV] pubdate: 2016-04-13  < storedate: 2016-04-26
Wed, 13 Apr 2016 16:16:05 +0000 is before store date of 2016-04-27.
  Ignoring search result as this is not the right issue.
Could not find Issue 6 of One-Punch Man (2014) using NZBGeek [api]
```

**Why the gate does not apply here.** The store-date comparison is a *periodical* tiebreak. An issue **number recycles** across runs — #6 of a 2016 run and #6 of a 1998 one are both "6" — so the street date is what says which run a result belongs to. A volume number does **not** recycle, and `manga_volume_satisfies()` already has to match it, so on a volume pass the date adds no disambiguation.

**And it actively subtracts.** Digital manga is routinely posted *before* the street date, and ComicVine store dates for manga volumes run weeks late, so the gate rejects the one correct release.

Across a 33-volume series this silently capped the library: every volume whose post happened to land after its recorded store date imported, and **the rest read as "not available"** — which looks like an indexer gap rather than a filter.

---

## 2. A trailing volume label is parsed into the series title

```
"VIZ Media - One-Punch Man Vol.33 (2026)"  ->  series: "VIZ Media-One-Punch Man Vol 33"   (matches nothing)
"VIZ Media - One-Punch Man v33 (2026)"     ->  series: "VIZ Media-One-Punch Man"          (correct)
```

The volume is read correctly in **both** cases — only the title differs.

**Cause:** two different things find the volume. The token walker finds the short form and truncates the title at the position it found it; the manga volume regex finds *both* forms but only overrides `series_volume`, and never touches the title. So the long form is detected, reported as the volume, and left in the series name at the same time.

The result is a release that parses to a series name matching no series, so a perfectly good file is never considered.

---

### Reproduce

1. Add a manga series with a populated volume ledger.
2. Search for a volume whose release was posted before its ComicVine store date -> rejected by the date gate (defect 1).
3. Search where the indexer names the release `... Vol.NN ...` rather than `... vNN ...` -> parses to an unmatchable series name (defect 2).

### Expected

- On a volume pass, the store-date tiebreak should not apply — the volume number already disambiguates.
- A trailing volume label that names the volume already detected belongs to the volume, not to the series title.

azuretek:
Fixes #820.

> **Builds on #815 and #816.** This branch carries both of those PRs' commits as prerequisites — the date gate lives in code #815 touches, and the parser change lives in code #816 touches. Only the **last two** commits are new here; once both merge, this diff reduces to those two.

### 1. `fix(manga): do not store-date gate a volume pass`

Take the same exit `UseFuzzy` already takes, **scoped to the volume pass** — `manga_match_name` is set only there, so a chapter pass keeps normal issue-number handling and keeps the tiebreak.

The reasoning is that the store date is a *periodical* disambiguator: issue numbers recycle across runs, so the street date says which run a result belongs to. Volume numbers do not recycle, and `manga_volume_satisfies()` already has to match the volume, so on a volume pass the date adds nothing — while actively rejecting correct releases, because digital manga is routinely posted before the street date and ComicVine store dates for manga volumes run weeks late.

Tests cover the fix and, **as controls**, that a periodical with identical dates and a manga series that is *not* on a volume pass both stay rejected. The gate is scoped, not disabled.

### 2. `fix(parse): a volume label is the volume, not part of the series title`

`strip_trailing_volume_label()` removes a trailing label **whose number is the volume that was actually detected**.

Requiring the numbers to agree is what keeps it safe. All of these are left alone:

- a series legitimately titled with a trailing number (`District 9`)
- a label naming a *different* volume than the one detected
- a label that is not at the end
- a title consisting of nothing but a label — stripping that would leave a name matching every series

## Test

| | |
|---|---|
| Full suite on this branch | **2889 passed, 8 failed, 1 skipped** |
| New tests, source reverted | **19 failed, 41 passed** |

The reverted run gives genuine assertion failures; the 41 that still pass include the scoping controls above, which assert *preserved* rejection behaviour and therefore correctly pass on both sides.

## Reading order

If you are reviewing the manga work as a set, this reads after #815 and #816 — those make a volume release findable and importable; this stops two later gates from discarding one that was found correctly.

### DIFF
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -77,25 +77,64 @@ def log_scan_summary(module, filename, candidate_count, selected_items, annual_c
     )
 
 
+_COLLECTED_TYPES = ("TPB", "HC", "GN")
+
+
+def numbered_by_volume(watch_values):
+    """Return whether this series names its files by volume rather than issue.
+
+    Collected editions and One-Shots have always been named this way, which is
+    why the scan exempts them from every check that assumes a file carries an
+    issue number: the weekly-pull cross-check, the "no issue number" rejection,
+    and the fallback that defaults a missing number to 1.
+
+    Manga belongs with them. In a periodical ``vNN`` names *which run* of the
+    series a release belongs to; in manga it names *which book*. A manga volume
+    file therefore carries no issue number at all, so applying the periodical
+    checks to it rejects a perfectly good file.
+
+    Rows built without the manga flag -- story arcs and one-offs assembled from
+    tables that do not carry it -- read as comics, which is the prior behaviour.
+    """
+    if watch_values.get("IsManga"):
+        return True
+    return watch_values["Type"] in _COLLECTED_TYPES + ("One-Shot",)
+
+
+def volume_match_settles_year(watch_values, volume_matched):
+    """Return whether a matched volume makes the filename's year irrelevant.
+
+    The year check exists because periodical issue numbers repeat across runs,
+    so ``Batman 1`` needs a year to say *which* Batman 1. A manga volume number
+    does not repeat -- there is one volume 1 -- so once it matches, the year
+    cannot be evidence of a wrong file.
+
+    It is routinely a mismatch, too: a provider dates the licensed English
+    printing while the release is labelled with the volume's original year. For
+    One-Punch Man v01 those are 2015 and 2014, and the file was rejected for it.
+
+    Deliberately manga-only. A collected edition's year mismatch can still mean
+    the wrong edition, so periodical and TPB behaviour is untouched.
+    """
+    return bool(volume_matched) and bool(watch_values.get("IsManga"))
+
+
 def volume_identifies_file(watch_values):
-    """Return whether this series numbers its files by volume rather than issue.
+    """Return whether a scanned file is located by volume rather than issue.
 
-    Collected editions (TPB/HC/GN spanning more than one entry) and One-Shots
-    have always been numbered this way, so a scanned file is located by the
-    ``vNN`` token in its name rather than by an issue number.
+    This is :func:`numbered_by_volume` qualified by the run length, because a
+    collected edition only numbers *files* by volume when the run actually has
+    volumes to distinguish: TPB/HC/GN spanning more than one entry, or a
+    One-Shot standing alone.
 
-    Manga joins them for a different reason. In a periodical, ``vNN`` names
-    *which run* of the series a release belongs to; in manga it names *which
-    book*, so the volume is the file's identity exactly as an issue number is
-    elsewhere. A manga volume file carries no issue number at all, so without
-    this the scan derives no number for it, matches it against every issue in
-    the series, and selects none of them.
+    Manga is not qualified that way. Volume 1 of a one-volume manga is still
+    identified by its volume, so the run length says nothing useful here.
     """
     if watch_values.get("IsManga"):
         return True
     series_type = watch_values["Type"]
     total = watch_values["Total"]
-    return (series_type in ("TPB", "HC", "GN") and total > 1) or (series_type == "One-Shot" and total == 1)
+    return (series_type in _COLLECTED_TYPES and total > 1) or (series_type == "One-Shot" and total == 1)
 
 
 def summarize_scan_matches(normal_items, arc_items):
@@ -1384,27 +1423,13 @@ class PostProcessor(object):
                             temploc = just_the_digits.replace("_", " ")
                             temploc = re.sub("[\\#']", "", temploc)
                         else:
-                            if any(
-                                [
-                                    cs["WatchValues"]["Type"] == "TPB",
-                                    cs["WatchValues"]["Type"] == "GN",
-                                    cs["WatchValues"]["Type"] == "HC",
-                                    cs["WatchValues"]["Type"] == "One-Shot",
-                                ]
-                            ):
+                            if numbered_by_volume(cs["WatchValues"]):
                                 temploc = "1"
                             else:
                                 temploc = None
                         datematch = "False"
 
-                        if temploc is None and all(
-                            [
-                                cs["WatchValues"]["Type"] != "TPB",
-                                cs["WatchValues"]["Type"] != "GN",
-                                cs["WatchValues"]["Type"] != "HC",
-                                cs["WatchValues"]["Type"] != "One-Shot",
-                            ]
-                        ):
+                        if temploc is None and not numbered_by_volume(cs["WatchValues"]):
                             logger.info(
                                 "this should have an issue number to match to this particular series: %s"
                                 % cs["ComicID"]
@@ -1662,14 +1687,9 @@ class PostProcessor(object):
                                             alts = x["AS_Alt"]
                                     alt_listing = [True if x.lower() == dynamic_seriesname else False for x in alts]
 
-                                    if any([cs["DynamicName"] == dynamic_seriesname, alt_listing]) and all(
-                                        [
-                                            cs["WatchValues"]["Type"] != "TPB",
-                                            cs["WatchValues"]["Type"] != "GN",
-                                            cs["WatchValues"]["Type"] != "HC",
-                                            cs["WatchValues"]["Type"] != "One-Shot",
-                                        ]
-                                    ):
+                                    if any(
+                                        [cs["DynamicName"] == dynamic_seriesname, alt_listing]
+                                    ) and not numbered_by_volume(cs["WatchValues"]):
                                         logger.fdebug(
                                             "name match exact : %s - %s" % (cs["DynamicName"], dynamic_seriesname)
                                         )
@@ -1798,15 +1818,7 @@ class PostProcessor(object):
                                 else:
                                     logger.fdebug("not a match")
 
-                                if all(
-                                    [
-                                        second_check is False,
-                                        cs["WatchValues"]["Type"] != "TPB",
-                                        cs["WatchValues"]["Type"] != "GN",
-                                        cs["WatchValues"]["Type"] != "HC",
-                                        cs["WatchValues"]["Type"] != "One-Shot",
-                                    ]
-                                ):
+                                if second_check is False and not numbered_by_volume(cs["WatchValues"]):
                                     logger.fdebug(
                                         "%s %s in filename don't match up to what's in the dB for %s [%s]. This is a wrong match. Continuing..."
                                         % (
@@ -1927,6 +1939,13 @@ class PostProcessor(object):
                                     logger.fdebug(
                                         "%s[LONE-VOLUME/NO YEAR][MATCH] Only Volume on watchlist matches, no year present in filename. Assuming match based on volume and title."
                                         % module
+                                    )
+                                    datematch = "True"
+
+                                if datematch == "False" and volume_match_settles_year(cs["WatchValues"], lonevol):
+                                    logger.fdebug(
+                                        "%s[MANGA][VOLUME MATCH] Volume %s matched, so the year in the filename (%s) does not decide this file. Assuming match based on volume and title."
+                                        % (module, watchmatch["series_volume"], watchmatch["issue_year"])
                                     )
                                     datematch = "True"
 
@@ -2080,6 +2099,7 @@ class PostProcessor(object):
                                     "Publisher": av["IssuePublisher"],
                                     "Total": int(av["TotalIssues"]),
                                     "Type": av["Type"],
+                                    "IsManga": series_kind.is_manga(av),
                                     "IsArc": True,
                                 },
                             }
@@ -2119,18 +2139,7 @@ class PostProcessor(object):
                                 nm += 1
                             else:
                                 try:
-                                    if (
-                                        any(
-                                            [
-                                                v[i]["WatchValues"]["Type"] == "TPB",
-                                                v[i]["WatchValues"]["Type"] == "GN",
-                                                v[i]["WatchValues"]["Type"] == "HC",
-                                            ]
-                                        )
-                                        and v[i]["WatchValues"]["Total"] > 1
-                                    ) or all(
-                                        [v[i]["WatchValues"]["Type"] == "One-Shot", v[i]["WatchValues"]["Total"] == 1]
-                                    ):
+                                    if volume_identifies_file(v[i]["WatchValues"]):
                                         if arcmatch["series_volume"] is not None:
                                             just_the_digits = re.sub("[^0-9]", "", arcmatch["series_volume"]).strip()
                                         else:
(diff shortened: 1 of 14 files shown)


azuretek:

azuretek:
Added a changeset here too — `.changeset/manga-volume-date-and-title.md`. You asked for one on #806 and #813, and every one of these is operator-visible, so I wrote the rest rather than have you ask six more times. Each is outcome-first per CONTRIBUTING — happy to reword any that reads wrong to you.

comicarr/postprocessor.py
```diff
@@ -77,6 +77,66 @@ def log_scan_summary(module, filename, candidate_count, selected_items, annual_c
...
+
+    This is :func:`numbered_by_volume` qualified by the run length, because a
+    collected edition only numbers *files* by volume when the run actually has
+    volumes to distinguish: TPB/HC/GN spanning more than one entry, or a
+    One-Shot standing alone.
+
+    Manga is not qualified that way. Volume 1 of a one-volume manga is still
+    identified by its volume, so the run length says nothing useful here.
+    """
+    if watch_values.get("IsManga"):
```
frankieramirez, line 199: A Manual Run will mark Chainsaw Man c181 as chapter 1 downloaded. FileChecker sets booktype manga, then defaults series_volume to v1 when it didn't find a volume token, and we look that up as Int_IssueNumber. v33 lands on chapter 33 for the same reason.

`_process_manga` already looks volume-only files up by VolumeNumber. This should be file-level like `is_volume_target` (a volume token and no chapter), and it shouldn't let that v1 default drive the lookup.

comicarr/postprocessor.py
```diff
@@ -1916,6 +1942,13 @@ def Process(self):
                                     )
                                     datematch = "True"
 
+                                if datematch == "False" and volume_match_settles_year(cs["WatchValues"], lonevol):
```
frankieramirez: `lonevol` is filename volume equals ComicVersion. MangaDex leaves ComicVersion unset, so it becomes 1, and One-Punch Man v06 with a 2014 filename year against a 2015 ledger year still gets rejected.

Once the scan lookup is file-level, pass that book-volume match in here instead of `lonevol`.

comicarr/search_filer.py
```diff
@@ -1122,6 +1173,16 @@ def _match_entry(self, entry, is_info):
...
+                    or all(
+                        [
+                            manga_volume_pass,
+                            pc_in is None,
+                            manga_volume_satisfies(
+                                filecomic.get("volume") or filecomic.get("series_volume"),
+                                findcomiciss,
+                            ),
+                        ]
+                    )
```
frankieramirez, line 1199: Nothing in the suite runs this arm. The store-date tests stub justthedigits as 1, so they accept via intIss == comintIss, and we could delete this block with CI still green.

Could we add an evaluate_entry case for a v06 title with no issue digits that should accept, and a v07 that should reject?

tests/unit/test_filechecker_volume_label.py
```diff
@@ -0,0 +1,95 @@
...
+    @staticmethod
+    def _source():
+        import inspect
+
+        from comicarr.filechecker import FileChecker
+
+        return inspect.getsource(FileChecker.parseit)
+
+    def test_parseit_strips_the_label_from_the_series_title(self):
+        assert "strip_trailing_volume_label(series_name, detected_volume)" in self._source(), (
```
frankieramirez: This only checks that the call string exists in parseit. If the return is discarded or the volume argument is wrong, CI still passes.

Can we run listFiles on something like `One-Punch Man Vol.33 (2014) (Digital).cbz` and assert series_name is One-Punch Man?

frankieramirez (commented):
Date-gate skip and the title strip look scoped correctly. Folder scan still looks volumes up as chapters, and that will write Have onto the wrong row.

comicarr/search_filer.py
```diff
@@ -463,8 +484,25 @@ def _match_entry(self, entry, is_info):
                     logger.fdebug("Invalid date found. Unable to continue - skipping result. Error returned: %s" % e)
                     _reject("invalid.pubdate_missing", cause=e)
 
-        if UseFuzzy == "1":
-            logger.fdebug("Year has been fuzzied for this series, ignoring store date comparison entirely.")
+        if UseFuzzy == "1" or manga_volume_pass:
```
frankieramirez, line 487: This skips the store-date gate. yearmatch still runs, and the vers4vol bypass is `len(series_volume) < 4`, which only covers v1-v99. mangaCheck passes ComicYear as the series year (One Piece 1997) while v100 is labeled 2021, so we reject it.

Setting yearmatch here the same way, or counting digits after the v, would let v100+ through.

comicarr/postprocessor.py
azuretek: Fixed in 48d8ac1, file-level as you asked.

`volume_identifies_file` now takes the parsed file and defers to `is_volume_target`, so the volume-vs-chapter rule has one owner instead of a second, weaker copy here:

```python
if watch_values.get("IsManga"):
    parsed = parsed_file or {}
    return is_volume_target(parsed.get("manga_chapter"), parsed.get("manga_volume"))
```

`matchIT` did not surface `manga_chapter`/`manga_volume` — `parseit` produces both but the match result dropped them — so those are carried through now. `series_volume` cannot stand in for them for exactly the reason you gave: its `v1` default makes a chapter file and a genuine volume 1 indistinguishable.

All three call sites (watchlist scan, arc scan, one-off scan) had the same shape and already had a matchIT result in hand, so all three now pass it and the fix is uniform. Omitting it returns False rather than True, so an unparsed file claims no volume instead of claiming volume 1.

Two of the existing tests asserted the old series-level answer and now assert the file-level one; added `c181` and the volume+chapter case (a MangaDex chapter names its containing volume, so it must read as a chapter).

comicarr/postprocessor.py
azuretek: Done in 48d8ac1 — now that the scan lookup is file-level, this is fed that result:

```python
book_volume_matched = volume_identifies_file(watch_values, watchmatch)
...
if datematch == "False" and volume_match_settles_year(cs["WatchValues"], book_volume_matched):
```

It is computed at the point `lonevol` is reset, inside `if datematch == "True"` — so reaching it means the number matched an issue row, and `volume_identifies_file` being True means that number came from the file's own volume token. That is the book-volume match, without ComicVersion in the path at all.

Added a parametrised test over v01/v06/v20/v100 with `ComicVersion: None`, which is the MangaDex shape — under `lonevol` only v01 passed. Plus a chapter control, since the widening has to stop at volumes.

comicarr/search_filer.py
azuretek: You're right that nothing ran it. Added `TestMangaVolumeAcceptanceArm` in 48d8ac1 with the two cases you asked for — v06 accepts, v07 rejects — plus v100 for the other thread.

Reaching the arm needed the parse stubbed as a volume release rather than as issue 1: `justthedigits=None` so `pc_in` is None, and `volume`/`series_volume` on the matchIT result so there is something for `manga_volume_satisfies` to compare. That is the shape the store-date tests could not produce.

Verified by deleting the arm: v06 and v100 fail, v07 still passes — which is the right asymmetry, since v07 must be rejected with or without it.

tests/unit/test_filechecker_volume_label.py
azuretek: You were right, and it was worse than a weak test — the behaviour was still broken.

Replaced the source assertions with real `listFiles` runs, and the first one failed:

```
'One-Punch Man Vol.33 (2014) (Digital).cbz' -> series_name='One-Punch Man Vol .33'
```

The call string was present, so the old test passed, but the strip never fired. The token walker rewrites `Vol.33` as `Vol .33` — the dot moves *after* the space — and `_TRAILING_VOLUME_LABEL` only allowed `\.?\s*`, dot first. So it matched `Vol 33` and `v33` and missed the long form it exists for. Now `\s*\.?\s*`, and `Vol.33` parses as `One-Punch Man`.

`TestVolumeLabelParsing` runs the parse for all five spellings, asserts the label never reaches the series title, and pins that the short and long forms of one release parse alike. A `Chainsaw Man 165.cbz` control keeps its issue number and claims no volume, so the strip cannot invent one.

The volume is compared through `volume_numbers_match` rather than by string: `v033` parses as `v033`, and normalising padding is the ledger's job, not the parser's.

Both new assertions fail against the old regex.

comicarr/search_filer.py
azuretek: Fixed by counting digits after the v, in 48d8ac1.

Traced `v100` through: `len("100") == 4` is false for the [Vxxxx] arm, `== 1` false for [Vx], and the third arm tested `len("v100") < 4` — 4 characters, so false — while the fourth needs a bare digit string. It matched **no arm**, so `fndcomicversion` stayed None and the bypass never ran.

The third arm was the odd one out: its two siblings measure `series_volume[1:]` and it measured the whole string. Now consistent:

```python
elif parsed_comic["series_volume"][1:].isdigit() and len(parsed_comic["series_volume"][1:]) < 4:
```

Chose this over setting `yearmatch` at the manga-pass site because it fixes the case for everything, not just manga — a `v100` collected edition hit the same hole — and it leaves the year/volume distinction where it already lives. Four digits after the v is still a year and is still claimed by [Vxxxx] first.

Covered by `test_a_three_digit_volume_is_still_read_as_a_volume`, which fails when the length test is put back on the whole string.

azuretek:

The replies on this PR cite `48d8ac1`, which was the head before the rebase. That commit is now `cfaf6bb`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/manga-volume-match-20260903`.

azuretek:
Rebased onto current main (`f8fc9b4`). **Six of the fourteen commits are already on main via #815 and have been dropped**, so this PR is now half the size it was:

- `f279461` fix(search): pass manga chapter/volume numbers from searchforissue
- `07764f7` fix(search): search manga volume targets without an issue-number suffix
- `8d43904` fix(search): match manga volume results against the real series name
- `f40c4b3` fix(search): accept a manga volume release on its volume number
- `dab3c5c` refactor(manga): give the volume rules one owner in the ledger
- `254387d` fix(search): read a manga release's vNN as the volume, not the series run

`_build_manga_search_terms`, `search_terms_for_target` and `volume_numbers_match` are all on main already, so these replayed only as whitespace conflicts against main's own formatting. Dropping them is what makes the rest reviewable: the only file this PR still touches that #815 owns is `comicarr/search_filer.py`, and only for the per-file volume-vs-chapter decision.

Old to new, for the eight that remain:

| before | after |
| --- | --- |
| `cfaf6bb` | `afac3f0` fix(manga): decide volume-vs-chapter per file, and stop losing v100 |
| `8d0a1c8` | `7c0ace4` chore: add changeset for the volume date gate and title parse fixes |
| `b1c6f95` | `704cb47` style: satisfy ruff check and ruff format on this branch |
| `1e11be6` | `62e0679` fix(parse): a volume label is the volume, not part of the series title |
| `fc97401` | `cdca671` fix(manga): do not store-date gate a volume pass |
| `275d6cb` | `372317c` fix(manga): give the folder scan one owner for volume-numbered series |
| `83a4a9a` | `877d7cc` fix(manga): number a watchlisted manga file by its volume during folder scan |
| `8f176f1` | `d7d10ba` fix(manga): parse volume/chapter tokens followed by release tags |

Pre-rebase head kept at `azuretek/comicarr@backup/manga-volume-match-20260905`.

Green after the rebase: 3015 passed in the unit suite, and `ruff check` plus `ruff format --check` clean on `comicarr/`.

azuretek:
Rebased onto current main (`c371ef5`), which this branch was conflicting with.

**#816 merging changed what this PR should contain.** Four of the ten commits are now on main through it and have been dropped, and a fifth has been reduced and renamed, so this branch no longer touches `postprocessor.py` at all. It is 5 files now, and every one of them is search or parse side.

The reduced commit was `fix(manga): decide volume-vs-chapter per file, and stop losing v100`. Two of its four fixes, the per-file volume-versus-chapter predicate and the year-settle that reads it, are fixed on main by a different route: `volume_identifies_file()` stays series-level and each call site is guarded with `and not chapter_named_file(match)`. That is merged, so both are dropped here. The `manga_chapter`/`manga_volume` carry-through in `filechecker.py` went with them, since it existed only to feed that predicate and now has no consumer.

What is left of it is the two fixes that nothing on main covers, and it is renamed to say so: `fix(search): stop losing v100, and strip the long-form volume label`.

So the PR is now exactly its title. Three defects, all in search and parse:

- the store-date gate rejecting a volume pass (`search_filer.py`)
- a volume label being read as part of the series title (`filechecker.py`)
- `v100` measured as 4 characters against the `< 4` version arm, so One Piece v100 labelled 2021 against a 1997 series year was rejected

Old to new:

| before | after | |
|---|---|---|
| `37db3bd` | dropped | parse volume/chapter tokens followed by release tags |
| `b95a779` | dropped | number a watchlisted manga file by its volume during folder scan |
| `cf3af08` | dropped | give the folder scan one owner for volume-numbered series |
| `9e3bfdb` | dropped | keep the folder series name for a tagged chapter-only file |
| `0ed9e06` | `3905719` | do not store-date gate a volume pass |
| `e2cbc82` | `2df3a50` | a volume label is the volume, not part of the series title |
| `e1da334` | `5055f99` | satisfy ruff check and ruff format on this branch |
| `3e1d1cb` | `9fea844` | add changeset for the volume date gate and title parse fixes |
| `47f8fc1` | `f834edf` | reduced to the v100 and long-form-label fixes, renamed and message rewritten |
| `7ad75a1` | `7a2cfaf` | house punctuation in the text this branch adds |

The four dropped ones were confirmed against main by patch-id, not by subject line. The pre-rebase head is kept at `azuretek/comicarr@backup/manga-volume-match-20260906`.

Green after the rebase, on the ruff version CI pins (0.15.7):

- `pytest tests/unit -m "not slow"`: 3035 passed, 0 failed. Main's baseline on the same command is 3006, so this adds 29 and removes none.
- `pytest tests/integration -m "not slow"`: 14 passed.
- `ruff check comicarr/`, `ruff format --check comicarr/`, `python scripts/run_ruff.py modern`: all clean.

PR #831, opened 2026-09-02
fix: accept the placement status post-processing writes to issues

### LINKED ISSUE #830: Imported one-off stranded at 'snatched' forever: recovery's issues done-signal tests a status that table never holds
## Summary

A one-off whose download completed **and was imported into the library** can be stranded at journal stage `snatched` permanently. Every restart re-classifies it, logs `UNKNOWN (downloader API unreachable / transient)`, and leaves it unchanged — while the library plainly shows the issue `Downloaded` with a `Location`.

The row never reaches a terminal stage, so it is invisible to `/api/attention`: no screen in the app shows it, and nothing ever closes it.

## Root cause

`comicarr/app/downloads/recovery_classify.py`:

```python
def _issue_post_processed(issueid):
    """True iff issues.Status == 'Post-Processed' for this IssueID — an
    authoritative "already completed" signal that survives history eviction."""
    ...
    return bool(rec) and rec["Status"] == "Post-Processed"
```

But post-processing writes those two statuses to **different tables**, in the same block — `comicarr/postprocessor.py` ~4293-4306:

```python
db.upsert(
    "issues",
    {"Status": "Downloaded", "Location": filename},
    {"IssueID": issueid},
)
...
db.upsert(
    "snatched",
    {"Status": "Post-Processed"},
    {"IssueID": issueid, "Status": "Snatched"},
)
```

`'Post-Processed'` is a **`snatched`**-table status. The `issues` table never receives it, so `_issue_post_processed()` can never return `True`.

Confirmed on a live instance — every issue in the library, 220 rows:

```
Downloaded   195
Wanted        18
Failed         7
```

`Post-Processed` appears zero times.

### Why that one dead predicate strands the row

`has_done_signal()` has three signals, and for a synthetic-HIGHCOUNT one-off the other two are unavailable by design:

```python
if _journal_stage_done(row):        # stage is still `snatched` -> False
    return True
if _issue_post_processed(issueid):  # can never be True
    return True
if journal.is_synthetic_oneoff(issueid):
    logger.fdebug(... "nzblog-presence is ADVISORY only ...")
    return False                    # <- one-offs stop here
```

So for a one-off at stage `snatched`, `_issue_post_processed()` is the **only** signal that can fire — and it cannot. `has_done_signal()` therefore always returns `False`, no matter how completely the work finished.

The same module already answers this exact question correctly elsewhere:

```python
_PLACED_STATUSES = ("Downloaded", "Post-Processed")

def _row_shows_placement(rec):
    if (rec["Status"] or "") in _PLACED_STATUSES:
        return True
```

Two helpers in one file disagree about what a completed issue looks like.

## Observed

Real log lines, repeating identically on every restart for the same two rows:

```
DEBUG   :: comicarr.has_done_signal.278 : [RECOVERY-CLASSIFY] one-off release_key=oneoff|nzbgeek|One-Punch.Man.v30.2025.Digital.LuCaZ|One-Punch.Man.v30.2025.Digital.LuCaZ — nzblog-presence is ADVISORY only; journal stage is authoritative.
WARNING :: comicarr._nzbget_history.373 : [RECOVERY-CLASSIFY] NZBGet row oneoff|nzbgeek|One-Punch.Man.v30.2025.Digital.LuCaZ|... has no NZBID to probe.
WARNING :: comicarr.classify_details.594 : [RECOVERY-CLASSIFY] oneoff|nzbgeek|One-Punch.Man.v30.2025.Digital.LuCaZ|... -> UNKNOWN (downloader API unreachable / transient) — journal stage left unchanged.
WARNING :: comicarr._resolve_row.818 : [RECOVERY] oneoff|nzbgeek|One-Punch.Man.v30.2025.Digital.LuCaZ|... -> UNKNOWN (transient downloader outage) — stage left UNCHANGED, reclassified next start.
```

Startup replay summary — the two rows are the `unknown-unchanged` pair:

```
[RECOVERY] Startup recovery replay complete: {... 'open': 33,
  'actions': {'unknown-unchanged': 2, 'downloaded-pp-enqueued': 31}}
```

Meanwhile the work was demonstrably finished:

| Evidence | Value |
|---|---|
| NZBGet history status | `SUCCESS/ALL`, health 1000 |
| File in library | `/…/Manga/One-Punch Man (2014)/One-Punch Man v30 (2025) (Digital) (LuCaZ).cbz` |
| `issues.Status` | `Downloaded` |
| `issues.Location` | `One-Punch Man v30 (2025) (Digital) (LuCaZ).cbz` |
| Series `Have` count | includes it (16/33) |
| `pipeline_journal.stage` | **`snatched`**, `status` NULL, 21h and counting |

## Reproduce

1. Grab a one-off whose `IssueID` is >= `HIGHCOUNT_FLOOR` (900000) — i.e. `journal.is_synthetic_oneoff()` is true.
2. Let the download complete and post-process successfully. `issues` now holds `Status='Downloaded'` + `Location`; `snatched` holds `Status='Post-Processed'`.
3. Ensure the journal row is still at stage `snatched` (its terminal write did not land — e.g. the NZBID was never recorded, so the probe cannot resolve it).
4. Restart Comicarr.
5. `has_done_signal()` returns `False`; the row is classified `UNKNOWN` and left unchanged. Repeat from 4 indefinitely.

## Why it survived

The existing regression test sets up a state the application never produces:

```python
def test_absent_but_issue_post_processed_is_complete_not_gone():
    with get_engine().begin() as conn:
        conn.execute(issues.insert().values(IssueID="I4", Status="Post-Processed"))
```

It writes `Post-Processed` **into `issues` directly**. Because production writes that value to `snatched` instead, the test passes while the production path can never take the same branch.

## Proposed fix

Accept the statuses that actually mean "placed" — reuse `_PLACED_STATUSES`, which the same module already uses for this question — and rename the predicate to `_issue_completed` so the name stops implying a status the table never holds.

Controls matter here: widening the check must not turn an in-flight or failed one-off into "done", or the fix becomes indistinguishable from deleting the check.

## Verified

| | `has_done_signal()` | replay action |
|---|---|---|
| Before | `False` | `unknown-unchanged`, forever |
| After | `True` | closes as done via the existing placement path |

Test suite on the branch: **2842 passed, 8 failed** — the same 8 failures present on pristine `main` (`tests/integration/test_schema_migration_dialects.py`), so zero new failures. `ruff check` and `ruff format --check` pass on both changed files; the repo's other 72 pre-existing ruff findings are untouched.

## Related, filed separately

`_nzbget_history()` maps "this row has no NZBID to probe" onto the **`unreachable`** state, which `classify_details()` reports as a *transient* outage and retries on every restart. A permanently missing identifier is not a transient outage, and NZBGet was healthy throughout. That is an independent bug and is not addressed by this change.

azuretek:
Fixes #830.

## Problem

`_issue_post_processed()` tested `issues.Status == 'Post-Processed'`. Post-processing writes that status to the **`snatched`** table and writes `'Downloaded'` + `Location` to **`issues`**, in the same block (`postprocessor.py` ~4293-4306):

```python
db.upsert("issues",   {"Status": "Downloaded", "Location": filename}, {"IssueID": issueid})
db.upsert("snatched", {"Status": "Post-Processed"}, {"IssueID": issueid, "Status": "Snatched"})
```

So an `issues` row never carries `'Post-Processed'`, and the predicate could never return `True`. On a live library all 220 issue rows are `Downloaded`/`Wanted`/`Failed`; `Post-Processed` appears zero times.

That is not a cosmetic dead branch. For a synthetic-HIGHCOUNT one-off it removes the **only** done-signal available: `has_done_signal()` returns early for one-offs because nzblog-presence is advisory for them, and a stranded row's stage is still `snatched`, so stage-rank cannot supply it either. A fully downloaded and imported one-off was therefore left at `snatched` permanently, re-probed and logged as `UNKNOWN (downloader API unreachable / transient)` on every restart while the library showed the issue `Downloaded` with a `Location`.

## Change

- Widen the predicate to `_PLACED_STATUSES` — the tuple this same module already uses for exactly this question in `_row_shows_placement()`. Two helpers in one file previously disagreed about what a completed issue looks like.
- Rename `_issue_post_processed` → `_issue_completed` (private, single call site) so the name no longer implies a status the table never holds.
- Comment the failure mode at the change site, and correct the module/`has_done_signal` docstrings that documented the old condition.

No behaviour change for any row that already had a done-signal.

## Test

`tests/unit/test_recovery_classify.py` gains the production shape: a synthetic-HIGHCOUNT one-off with `issues.Status='Downloaded'` + `Location`, its **nzblog row deliberately left present** and stage left at `snatched`, so the issues test is provably the only signal in play.

It fails on unpatched `main` with the production symptom:

```
________________ test_oneoff_issue_downloaded_is_a_done_signal _________________
tests/unit/test_recovery_classify.py:630: in test_oneoff_issue_downloaded_is_a_done_signal
    assert recovery_classify.has_done_signal(row) is True
E   AssertionError: assert False is True
E    +  where False = <function has_done_signal ...>({'release_key': 'oneoff|nzbgeek|One-Punch.Man.v30.2025.Digital.LuCaZ|...', 'issueid': '1099554', ...})

1 failed, 43 passed
```

and passes with the fix (`44 passed`).

Controls included so the change is distinguishable from simply removing the check — these must stay negative, and do so in both directions:

- `Wanted`, `Snatched`, `Failed` → still not a done-signal, still classify `STILL`
- no `issues` row at all → still not a done-signal
- the pre-existing `Post-Processed` acceptance → still a done-signal (no regression)

## Verification

| | Result |
|---|---|
| Full suite, this branch | **2842 passed**, 8 failed, 1 skipped |
| Full suite, pristine `main` | 2836 passed, **8 failed**, 1 skipped |
| Delta | +6 passing (the new tests), **0 new failures** |

The 8 failures are pre-existing on `main` (`tests/integration/test_schema_migration_dialects.py`) and unrelated to this change.

`ruff check` and `ruff format --check` both pass on the two changed files. The repo's other 72 pre-existing ruff findings and 32 unformatted files are left untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery detection for completed downloads by recognizing both “Downloaded” and “Post-Processed” statuses.
  * Prevented items with incomplete or missing placement information from being incorrectly marked as complete.

* **Tests**
  * Added coverage for completed, incomplete, failed, and missing download status scenarios.

### DIFF
--- a/comicarr/app/downloads/recovery_classify.py
+++ b/comicarr/app/downloads/recovery_classify.py
@@ -23,8 +23,8 @@ STILL/COMPLETE/UNKNOWN.
 history is finite and operator/auto-pruned: a release that completed and was
 post-processed while the app was down, then had its history row evicted, reads
 as absent. Before classifying any "absent" as GONE we cross-check the
-authoritative done-signals (issues.Status == 'Post-Processed' / nzblog row
-absent / journal stage already post_processing+).
+authoritative done-signals (issues.Status shows placement / nzblog row absent
+/ journal stage already post_processing+).
 
 One-off caveat: synthetic-HIGHCOUNT IssueID one-offs have a non-persisted
 IssueID that diverges across restart, and nzblog() has a mid-flight
@@ -63,9 +63,25 @@ def _journal_stage_done(row):
     return rank is not None and pp_rank is not None and rank >= pp_rank
 
 
-def _issue_post_processed(issueid):
-    """True iff issues.Status == 'Post-Processed' for this IssueID — an
-    authoritative "already completed" signal that survives history eviction."""
+def _issue_completed(issueid):
+    """True iff the issues row shows this obligation already completed — an
+    authoritative "already done" signal that survives history eviction.
+
+    Accepts every status that means the file was placed (``_PLACED_STATUSES``),
+    not only 'Post-Processed'. Post-processing writes BOTH statuses in the same
+    block, to DIFFERENT tables (postprocessor ~4293-4306): 'Downloaded' +
+    Location onto *issues*, and 'Post-Processed' onto *snatched*. An issues row
+    therefore never carries 'Post-Processed', so testing for it here could
+    never return True.
+
+    That matters because this is the ONLY done-signal a synthetic-HIGHCOUNT
+    one-off can have: nzblog-presence is advisory for those (see
+    has_done_signal) and a stranded row's stage is still `snatched`. A fully
+    downloaded and imported one-off was therefore stranded at `snatched`
+    forever — each restart re-probed it and logged "UNKNOWN (downloader API
+    unreachable / transient)" while the library plainly showed the issue
+    Downloaded with a Location.
+    """
     if issueid is None:
         return False
     try:
@@ -73,7 +89,7 @@ def _issue_post_processed(issueid):
     except Exception as e:
         logger.warn("[RECOVERY-CLASSIFY] issues.Status lookup failed for %s: %s" % (issueid, e))
         return False
-    return bool(rec) and rec["Status"] == "Post-Processed"
+    return bool(rec) and (rec["Status"] or "") in _PLACED_STATUSES
 
 
 def _nzblog_present(issueid, provider, story_arc=None):
@@ -254,7 +270,7 @@ def has_done_signal(row):
 
     Returns True iff the release is authoritatively already complete:
       * journal stage is post_processing+, OR
-      * issues.Status == 'Post-Processed', OR
+      * the issues row shows placement (see _issue_completed), OR
       * nzblog row absent (deleted on PP success).
 
     One-off rule: for a synthetic-HIGHCOUNT one-off the nzblog-presence test
@@ -262,7 +278,9 @@ def has_done_signal(row):
     delete-reupsert window). For those rows the journal release_key/stage is
     AUTHORITATIVE and nzblog-presence is ADVISORY only — so we do NOT treat
     nzblog-absence as a done-signal for one-offs (an in-flight one-off must
-    not be misread as done/GONE).
+    not be misread as done/GONE). That rule makes _issue_completed the only
+    done-signal such a row can have while its stage is still `snatched`, so
+    that test must recognise the status post-processing actually writes.
     """
     row = row or {}
     if _journal_stage_done(row):
@@ -271,7 +289,7 @@ def has_done_signal(row):
     issueid = row.get("issueid")
     provider = row.get("provider")
 
-    if _issue_post_processed(issueid):
+    if _issue_completed(issueid):
         return True
 
     if journal.is_synthetic_oneoff(issueid):
--- a/comicarr/app/downloads/recovery_classify.py
+++ b/comicarr/app/downloads/recovery_classify.py
@@ -23,8 +23,8 @@ STILL/COMPLETE/UNKNOWN.
 history is finite and operator/auto-pruned: a release that completed and was
 post-processed while the app was down, then had its history row evicted, reads
 as absent. Before classifying any "absent" as GONE we cross-check the
-authoritative done-signals (issues.Status shows placement / nzblog row absent
-/ journal stage already post_processing+).
+authoritative done-signals (a library row's Status shows placement / nzblog
+row absent / journal stage already post_processing+).
 
 One-off caveat: synthetic-HIGHCOUNT IssueID one-offs have a non-persisted
 IssueID that diverges across restart, and nzblog() has a mid-flight
@@ -63,16 +63,16 @@ def _journal_stage_done(row):
     return rank is not None and pp_rank is not None and rank >= pp_rank
 
 
-def _issue_completed(issueid):
-    """True iff the issues row shows this obligation already completed — an
+def _library_status_done(issueid, story_arc=None):
+    """True iff a library row shows this obligation already completed — an
     authoritative "already done" signal that survives history eviction.
 
     Accepts every status that means the file was placed (``_PLACED_STATUSES``),
     not only 'Post-Processed'. Post-processing writes BOTH statuses in the same
     block, to DIFFERENT tables (postprocessor ~4293-4306): 'Downloaded' +
-    Location onto *issues*, and 'Post-Processed' onto *snatched*. An issues row
-    therefore never carries 'Post-Processed', so testing for it here could
-    never return True.
+    Location onto the *library* row, and 'Post-Processed' onto *snatched*. A
+    library row therefore never carries 'Post-Processed', so testing for it
+    here could never return True.
 
     That matters because this is the ONLY done-signal a synthetic-HIGHCOUNT
     one-off can have: nzblog-presence is advisory for those (see
@@ -81,15 +81,27 @@ def _issue_completed(issueid):
     forever — each restart re-probed it and logged "UNKNOWN (downloader API
     unreachable / transient)" while the library plainly showed the issue
     Downloaded with a Location.
+
+    Reads every table that can carry the row, via ``_library_rows``, rather
+    than only *issues*. An annual's completion is written onto *annuals*
+    (updater.foundsearch, mode='want_ann') and an arc's onto *storyarcs*, so an
+    issues-only lookup missed both. Annuals are the case that bites: their
+    IssueIDs are real ComicVine ids, which routinely exceed the 900000
+    synthetic-one-off floor, so ``is_synthetic_oneoff`` reads True for them and
+    suppresses the nzblog fallback below — leaving this the only signal they
+    have, and it was looking in the wrong table. A successfully imported
+    annual classified GONE on every restart.
+
+    Status only, deliberately: a non-empty Location is placement evidence and
+    belongs to has_library_placement, which answers a different question. A
+    lookup failure returns False — never fabricate a done-signal.
     """
     if issueid is None:
         return False
-    try:
-        rec = db.select_one(select(issues.c.Status).where(issues.c.IssueID == str(issueid)))
-    except Exception as e:
-        logger.warn("[RECOVERY-CLASSIFY] issues.Status lookup failed for %s: %s" % (issueid, e))
+    recs = _library_rows(issueid, story_arc)
+    if not recs:
         return False
-    return bool(rec) and (rec["Status"] or "") in _PLACED_STATUSES
+    return any((rec["Status"] or "") in _PLACED_STATUSES for rec in recs)
 
 
 def _nzblog_present(issueid, provider, story_arc=None):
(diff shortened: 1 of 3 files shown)


tests/unit/test_recovery_classify.py
```diff
@@ -589,3 +589,74 @@ def test_reopen_candidate_story_arc_scoped_to_storyarcs_row():
...
+
+
+@pytest.mark.parametrize("status", ["Wanted", "Snatched", "Failed"])
+def test_oneoff_issue_not_placed_is_not_a_done_signal(status):
+    """Widening the accepted statuses must not turn an in-flight or failed
+    one-off into "done" — otherwise the fix is indistinguishable from
+    deleting the check."""
+    row = _imported_oneoff(status=status, location=None)
+    assert recovery_classify.has_done_signal(row) is False
+    assert recovery_classify.classify(row, probes=_probe("still")) == recovery_classify.STILL
```
frankieramirez: A miss that treated Snatched as placed would still pass this classify line, because still returns before has_done_signal runs. The assertion above is the real control. classify(absent) == GONE would match the happy-path test and actually catch that.

comicarr/app/downloads/recovery_classify.py
```diff
@@ -63,17 +63,33 @@ def _journal_stage_done(row):
...
+    """
     if issueid is None:
         return False
     try:
         rec = db.select_one(select(issues.c.Status).where(issues.c.IssueID == str(issueid)))
     except Exception as e:
         logger.warn("[RECOVERY-CLASSIFY] issues.Status lookup failed for %s: %s" % (issueid, e))
         return False
-    return bool(rec) and rec["Status"] == "Post-Processed"
+    return bool(rec) and (rec["Status"] or "") in _PLACED_STATUSES
```
frankieramirez: want_ann writes Downloaded onto annuals, so a HIGHCOUNT annual (IssueID >= 900000) still classifies GONE after a successful import. nzblog is skipped for those IDs and this lookup never sees the annuals row. _library_rows already walks annuals for placement. Can we reuse that?

azuretek: Fixed in de514ba — you were right, and it was worse than the annuals table alone.

Annual IssueIDs are real ComicVine ids, and those routinely sit above 900000, so `is_synthetic_oneoff` reads True for an ordinary annual. That suppresses the nzblog fallback, which leaves this lookup as the annual's *only* done-signal — and it was querying the wrong table. Every restart re-probed a cleanly imported annual and failed it.

Reused `_library_rows` as you suggested, so there is one owner for "which rows can carry evidence for this obligation", including the arc scoping:

```python
def _library_status_done(issueid, story_arc=None):
    if issueid is None:
        return False
    recs = _library_rows(issueid, story_arc)
    if not recs:
        return False
    return any((rec["Status"] or "") in _PLACED_STATUSES for rec in recs)
```

Kept it status-only rather than calling `has_library_placement`: a non-empty Location is placement evidence and that function answers a different question, so folding them together would have widened the done-signal past what this path should accept.

`has_done_signal` now reads the arc signal once, via a new `_row_story_arc`, and passes it down. That also removed the second inline copy of the payload parse that was sitting further down the same function.

Renamed to `_library_status_done` since "issue" was the wrong noun once it walks all three tables.

Added `test_imported_annual_is_a_done_signal` plus not-placed controls. The happy-path one fails without the source change (`has_done_signal` returns False), so it is a real control rather than a description.

comicarr/app/downloads/recovery_classify.py
```diff
@@ -63,17 +63,33 @@ def _journal_stage_done(row):
...
+    has_done_signal) and a stranded row's stage is still `snatched`. A fully
+    downloaded and imported one-off was therefore stranded at `snatched`
+    forever — each restart re-probed it and logged "UNKNOWN (downloader API
+    unreachable / transient)" while the library plainly showed the issue
+    Downloaded with a Location.
+    """
     if issueid is None:
         return False
     try:
         rec = db.select_one(select(issues.c.Status).where(issues.c.IssueID == str(issueid)))
```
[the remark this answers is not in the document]
azuretek: Fixed in de514ba, though not by editing that line — it is gone.

Per your other comment this lookup now goes through `_library_rows`, which already walks issues/annuals/storyarcs with the right arc scoping. That deleted the `db.select_one` call, the `try`/`except` around it and the `logger.warn` inside it. So both of these threads resolve as "the code no longer exists" rather than as a style change.

For the record on the guideline itself, in case it comes up on a future PR: `db.select_one` **is** the seam in this layer, not a bypass of one. It is defined in `comicarr/db.py`, takes a SQLAlchemy Core statement and parameterises through the engine, and it is what every sibling call in `comicarr/app/downloads/` uses — 6 more in this file alone and 8 in `recovery.py`. `db.DBConnection().action(...)` is the raw-SQL path used by the legacy `comicarr/*.py` modules. Switching this one call to it would have moved the module *away* from its own convention and traded generated parameter binding for hand-written SQL.

comicarr/app/downloads/recovery_classify.py
```diff
@@ -63,17 +63,33 @@ def _journal_stage_done(row):
...
     except Exception as e:
         logger.warn("[RECOVERY-CLASSIFY] issues.Status lookup failed for %s: %s" % (issueid, e))
```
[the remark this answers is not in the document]
azuretek: See the reply on the `logger.warn` thread — the enclosing lookup was replaced with `_library_rows`, so this line no longer exists in de514ba.

Noting the convention question for future PRs, since it will come up again: `logger.warn` is what this module and this package use — 11 calls in this file, 52 across `comicarr/app/downloads/`, and the surrounding probe failures all log at that level. It is exported from `comicarr/logger.py` alongside `error`. This particular line was also not introduced by this PR; the diff only moved it as part of the rename.

Changing one of eleven to `logger.error` would have made the file inconsistent with itself and promoted a recoverable lookup failure — which returns False and falls through to the other done-signals — to the same level as an actual error.

tests/unit/test_recovery_classify.py
azuretek: Good catch, and agreed the assertion was doing nothing. Fixed in de514ba.

`_probe("still")` returns STILL before `has_done_signal` is reached at all, so that line passed whether or not the status was read correctly — it was asserting the probe seam works, not the change. Switched to your suggestion:

```python
assert recovery_classify.has_done_signal(row) is False
assert recovery_classify.classify(row, probes=_probe("absent")) == recovery_classify.GONE
```

which now mirrors the happy path's `_probe("absent") == COMPLETE` exactly, so the pair of tests differ only in the thing under test.

Used the same shape for the new annual controls rather than repeating the mistake there.

azuretek:

The replies on this PR cite `de514ba`, which was the head before the rebase. That commit is now `b210548`; the sequence is otherwise identical, and the pre-rebase head is kept at `azuretek/comicarr@backup/recovery-issue-done-signal-20260903`.

Issue #734, opened 2026-08-18
Startup recovery marks downloads post_processed without running file placement


Rocketplanner83:
## Summary

Comicarr startup recovery can reconstruct missing snatched journal rows from legacy `Status`/`nzblog` state, mark them as complete, and advance them directly to `post_processed` **without invoking the normal file-placement path**.

This results in activity showing `import / succeeded` even though the downloaded comic files remain in the download directory and are never moved into the library.

## Environment

Comicarr is running as:

```text
python /opt/comicarr/Comicarr.py --nolaunch --datadir /config/comicarr
```

Effective identity:

```text
uid=568(comicarr) gid=568(comicarr) groups=568(comicarr)
```

Relevant configuration:

```ini
destination_dir = /books/comics
newcom_dir = /books/downloads/usenet
post_processing = True
file_opts = move
move_files = False
```

In the installed version, `move_files` appears to be a stale/unused configuration key. The runtime post-processing path uses `file_opts`, and `file_opts = move` is the operative setting.

## Filesystem and permission validation

Source and destination are on the same ZFS filesystem/dataset:

```text
<media-dataset>  zfs  /books/downloads/usenet
<media-dataset>  zfs  /books/comics
```

Container mounts are read-write:

```text
<host-media-path>/downloads/usenet -> /books/downloads/usenet rw=true
<host-media-path>/comics           -> /books/comics rw=true
```

Example affected release directory:

```text
<host-user> <container-group> 1000 568 775 <host-media-path>/downloads/usenet/<release-name>
```

Because Comicarr runs with GID `568`, mode `775` gives it group `rwx` access.

A non-destructive write test inside the Comicarr container as `568:568` succeeded:

```text
create temporary zero-byte file: PASS
delete temporary file: PASS
```

So this does not appear to be a permissions, umask, mount, or filesystem problem.

## Observed behavior

For affected releases, startup recovery logged:

```text
[RECOVERY] reconstructed missing snatched journal row
```

followed by:

```text
[RECOVERY] <issue-id>|<indexer> authoritatively done (Status/nzblog) — mark_done, skip.
```

The same sequence occurred for multiple affected issues.

After this, the pipeline journal showed:

```text
pipeline stage: post_processed
issue status: Snatched
library location: None
```

Activity later showed:

```text
import / succeeded
```

However:

* files remained under `/books/downloads/usenet`
* destination library folders remained empty
* there were no corresponding `importresults`
* there were no exception records
* there was no destination-path log
[...]

PR #833, opened 2026-09-02
fix: Treat a missing NZBGet NZBID as unprobeable, not unreachable

### LINKED ISSUE #832: Row with no NZBID is reported as a transient downloader outage and retried on every restart
## Summary

`_nzbget_history()` returns the **`unreachable`** probe state when a journal row simply has no `NZBID` recorded. `classify_details()` interprets `unreachable` as a transient downloader outage, returns `UNKNOWN`, and leaves the row unchanged "to be reclassified next start".

But a missing `NZBID` is a **permanent property of the row**, not a transient condition. No amount of retrying will make one appear, so the row is re-probed and re-logged on every single restart, forever, and the log accuses a downloader that is perfectly healthy.

## Root cause

`comicarr/app/downloads/recovery_classify.py`:

```python
def _nzbget_history(row, payload=None):
    """NZBGet: history lookup by NZBID. Reuses nzbget.NZBGet.historycheck()."""
    ...
    nzbid = payload.get("NZBID") or di.get("NZBID") or row.get("NZBID")
    if not nzbid:
        logger.warn("[RECOVERY-CLASSIFY] NZBGet row %s has no NZBID to probe." % row.get("release_key"))
        return "unreachable"          # <-- permanent condition, transient state
    try:
        ...
    except Exception as e:
        logger.warn("[RECOVERY-CLASSIFY] NZBGet unreachable probing %s: %s" % (nzbid, e))
        return "unreachable"          # <-- genuinely transient
```

Both branches return the same state, but they mean opposite things: the second is "ask again later", the first is "there is nothing here to ask about, ever".

Downstream, `classify_details()` cannot tell them apart:

```python
if raw_state == "unreachable":
    logger.warn(
        "[RECOVERY-CLASSIFY] %s -> UNKNOWN (downloader API unreachable / "
        "transient) — journal stage left unchanged." % rkey
    )
```

Note this early-return happens **before** the `has_done_signal()` cross-check, so a row that takes this path never gets the history-eviction guard applied to it at all.

## Observed

NZBGet was healthy and reachable throughout — the same replay successfully probed and closed 31 other rows in the same pass:

```
WARNING :: comicarr._nzbget_history.373 : [RECOVERY-CLASSIFY] NZBGet row oneoff|nzbgeek||0db5cc8c289a08a0 has no NZBID to probe.
WARNING :: comicarr.classify_details.594 : [RECOVERY-CLASSIFY] oneoff|nzbgeek||0db5cc8c289a08a0 -> UNKNOWN (downloader API unreachable / transient) — journal stage left unchanged.
WARNING :: comicarr._resolve_row.818 : [RECOVERY] oneoff|nzbgeek||0db5cc8c289a08a0 -> UNKNOWN (transient downloader outage) — stage left UNCHANGED, reclassified next start.
```

Identical output at 12:47, 13:32, 13:54, 14:26 … on each restart, for the same rows.

Note the release key: `oneoff|nzbgeek||0db5cc8c289a08a0` — the nzbname segment is empty, and the payload carries no `nzbname` and no `NZBID`. There is nothing in the row to probe *with*.

## Reproduce

1. Create a journal row whose payload has neither `NZBID` nor `download_info.NZBID` (e.g. a one-off grab where the snatch seam did not record the id).
2. Restart with NZBGet configured and healthy.
3. The row logs `has no NZBID to probe` → `UNKNOWN (downloader API unreachable / transient)` and is left unchanged.
4. Restart again. Identical output. The row never resolves.

## Why this matters beyond the noise

- **The diagnosis is wrong and misleads operators.** "Downloader API unreachable / transient" points at NZBGet, which is fine. Time gets spent checking a healthy service.
- **The row skips the eviction guard.** Because `unreachable` returns before `has_done_signal()`, a row with no NZBID is never cross-checked against the authoritative done-signals — even when the library can prove the work finished.
- **It is unbounded.** There is no attempt counter and no age cutoff, so a row in this state is retried at every startup for the lifetime of the install.

## Suggested direction

Distinguish "cannot probe this row" from "cannot reach the client" — e.g. a separate `unprobeable` raw state that falls through to the `has_done_signal()` cross-check rather than short-circuiting to `UNKNOWN`. That would let a row whose library placement is provable be closed correctly, and leave a genuinely unknown row unknown, without accusing a healthy downloader.

I have not opened a PR for this one because the right verdict for an unprobeable row is a design decision rather than an obvious substitution — happy to implement whichever direction you prefer.

## Related

Found alongside #830, which fixes the other half of why these particular rows were stranded. The two are independent: #830 restores the done-signal, this one is about the probe state being mislabelled.

frankieramirez:
## Summary

Fixes #832. A journal row with no `NZBID` was classified as a transient NZBGet outage (`unreachable` → `UNKNOWN`) and re-probed on every restart, even when NZBGet was healthy and other rows closed fine.

A missing NZBID is a permanent property of the row, not a reachability failure. The missing-id branch now returns a separate `unprobeable` raw state that falls through to the `has_done_signal()` history-eviction guard instead of short-circuiting to UNKNOWN. Genuine NZBGet API/exception failures stay `unreachable`.

## Changes

- `_nzbget_history()` returns `unprobeable` when the row has no NZBID, and still returns `unreachable` when `historycheck()` fails.
- `classify_details()` sends `unprobeable` through the done-signal cross-check: done-signal → `COMPLETE`; otherwise `UNKNOWN` (not `GONE`, and not "downloader API unreachable").
- Recovery replay logs "row has no client id to probe" for that UNKNOWN path instead of "transient downloader outage".
- Unit coverage for the built-in NZBGet probe, the injectable `unprobeable` seam, and the replay actions.
- Review follow-up: a replay that calls the real `_nzbget_history` with an empty payload (no `probes=` seam) so a reconstructed NZBGet row with no NZBID and no nzblog stays pinned to the same COMPLETE / PP-enqueue path as `absent`. The #734 `unreachable` → `manual_review` test remains the transient-outage case only.

Related #830 restores the done-signal for imported one-offs and is not folded in here.

## Release Impact

- [x] User-visible app change; changeset added with `npm run changeset` (operator-facing, outcome-first prose — see CONTRIBUTING.md)
- [ ] No app release impact; no changeset needed (omit rather than "No user-visible behaviour changes.")

## Testing

- [x] Tests pass locally (`pytest tests/unit -v`) — `tests/unit/test_recovery_classify.py` + `tests/unit/test_recovery_replay.py`: **96 passed**
- [ ] Frontend builds (`cd frontend && npm run build`) — N/A, backend-only
- [x] Backend lint/format (`ruff check comicarr/` and `ruff format --check comicarr/`) — `lint:modern` + ruff check/format on the changed files
- [ ] Frontend lint/format (`cd frontend && npm run lint && npm run format:check`) — N/A, backend-only
- [x] Contributor guards (`lint:guards` subset: retired globals, fail_reason registry, upsert tables, attention seam)
- [ ] Or one-shot: `npm run lint` (from repo root; requires `uv` + `frontend/node_modules`)
- [x] Manually tested (describe what you tested): exercised the built-in `_nzbget_history` missing-NZBID and API-failure paths plus replay of `unprobeable` with and without a done-signal, including a no-probes empty-payload replay (no live NZBGet)

## Screenshots

N/A — backend recovery classification only.

## Additional Notes

Unprobeable without a done-signal still returns UNKNOWN and is reclassified next start. That is intentional: we cannot ask the client and we cannot prove completion, so we must not mark the row `GONE`. The diagnosis no longer accuses a healthy NZBGet, and rows that *can* prove completion now close.

A reconstructed NZBGet row with no NZBID and no nzblog is treated like history-evicted `absent`: marked `downloaded` and put on `PP_QUEUE` even when the folder is unknown. Post-process validation owns the missing-folder quarantine. That is distinct from a genuine `unreachable` probe, which still takes the #734 `manual_review` path.

<div><a href="https://cursor.com/agents/bc-0c8eab4f-b266-4e8a-96ca-7c19eacb86ac?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0c8eab4f-b266-4e8a-96ca-7c19eacb86ac&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup recovery for downloads without a recorded NZBGet ID.
  * Completed downloads are now recognized when the library confirms completion.
  * Incomplete or unverifiable downloads remain marked as unknown without incorrectly reporting an NZBGet outage.
  * Genuine NZBGet connection failures continue to be retried as temporary outages.

* **Tests**
  * Added coverage for recovery behavior with missing download IDs, completed items, and connection failures.

### DIFF
--- a/comicarr/app/downloads/recovery_classify.py
+++ b/comicarr/app/downloads/recovery_classify.py
@@ -623,10 +623,7 @@ def classify_details(row, probes=None, payload=None):
 
     if has_done_signal(row):
         if raw_state == "unprobeable":
-            logger.fdebug(
-                "[RECOVERY-CLASSIFY] %s unprobeable BUT done-signal present "
-                "-> complete, NOT unknown." % rkey
-            )
+            logger.fdebug("[RECOVERY-CLASSIFY] %s unprobeable BUT done-signal present -> complete, NOT unknown." % rkey)
         else:
             logger.fdebug(
                 "[RECOVERY-CLASSIFY] %s absent in client BUT done-signal present "
@@ -637,8 +634,7 @@ def classify_details(row, probes=None, payload=None):
 
     if raw_state == "unprobeable":
         logger.warn(
-            "[RECOVERY-CLASSIFY] %s -> UNKNOWN (row has no client id to probe) "
-            "— journal stage left unchanged." % rkey
+            "[RECOVERY-CLASSIFY] %s -> UNKNOWN (row has no client id to probe) — journal stage left unchanged." % rkey
         )
         details["verdict"] = UNKNOWN
         return details
--- a/comicarr/app/downloads/recovery_classify.py
+++ b/comicarr/app/downloads/recovery_classify.py
@@ -364,14 +364,19 @@ def _sab_history_or_queue(row, payload=None):
 
 
 def _nzbget_history(row, payload=None):
-    """NZBGet: history lookup by NZBID. Reuses nzbget.NZBGet.historycheck()."""
+    """NZBGet: history lookup by NZBID. Reuses nzbget.NZBGet.historycheck().
+
+    A missing NZBID is a permanent property of the row (nothing to ask the
+    client), so it is ``unprobeable`` — not a transient ``unreachable``.
+    Genuine historycheck failures stay ``unreachable`` (#832).
+    """
     payload = payload if payload is not None else journal.load_payload(row.get("payload_json"))
     payload = payload or {}
     di = payload.get("download_info") or {}
     nzbid = payload.get("NZBID") or di.get("NZBID") or row.get("NZBID")
     if not nzbid:
         logger.warn("[RECOVERY-CLASSIFY] NZBGet row %s has no NZBID to probe." % row.get("release_key"))
-        return "unreachable"
+        return "unprobeable"
     try:
         from comicarr import nzbget
 
@@ -406,7 +411,7 @@ def _nzstat_to_raw(nzstat):
     return "complete"
 
 
-_RAW_PROBE_STATES = frozenset(("still", "complete", "absent", "unreachable"))
+_RAW_PROBE_STATES = frozenset(("still", "complete", "absent", "unreachable", "unprobeable"))
 
 
 def _probe_evidence(raw):
@@ -432,8 +437,14 @@ def _probe_evidence(raw):
     return raw_state, location, name, failed
 
 
-def _empty_details(verdict=UNKNOWN):
-    return {"verdict": verdict, "location": None, "name": None, "failed": None}
+def _empty_details(verdict=UNKNOWN, raw_state=None):
+    return {
+        "verdict": verdict,
+        "location": None,
+        "name": None,
+        "failed": None,
+        "raw_state": raw_state,
+    }
 
 
 def _probe_nzb(row, payload=None):
@@ -542,15 +553,21 @@ def _resolve_downloader(row, payload=None):
 def classify_details(row, probes=None, payload=None):
     """Classify one open journal row and return completion evidence.
 
-    Returns ``{verdict, location, name, failed}``. ``verdict`` is one of
-    STILL/COMPLETE/GONE/UNKNOWN. ``location``/``name``/``failed`` come from a
-    richer SAB/NZBGet probe result when present; they are None when the probe
-    returned only a raw state string. PURE: never mutates the journal.
+    Returns ``{verdict, location, name, failed, raw_state}``. ``verdict`` is
+    one of STILL/COMPLETE/GONE/UNKNOWN. ``location``/``name``/``failed`` come
+    from a richer SAB/NZBGet probe result when present; they are None when
+    the probe returned only a raw state string. ``raw_state`` is the
+    normalized probe vocabulary (including ``unprobeable``) so callers can
+    distinguish a missing client id from a transient outage. PURE: never
+    mutates the journal.
 
     `probes` (test seam): optional {downloader_type: callable(row)->raw}
     overriding the real client-query paths. `raw` is one of
-    "still"/"complete"/"absent"/"unreachable", or a historycheck-shaped dict
-    with ``status`` plus optional ``location``/``name``/``failed``.
+    "still"/"complete"/"absent"/"unreachable"/"unprobeable", or a
+    historycheck-shaped dict with ``status`` plus optional
+    ``location``/``name``/``failed``. ``unprobeable`` means this row has
+    nothing to ask the client (e.g. no NZBID); it is not a transient
+    outage and still receives the done-signal / history-eviction guard.
 
     `payload` (efficiency): the already-decoded payload_json dict, parsed once
     per replay row by the caller and threaded in to avoid re-parsing it for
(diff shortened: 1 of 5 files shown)


comicarr/app/downloads/recovery_classify.py
```diff
@@ -599,13 +622,23 @@ def classify_details(row, probes=None, payload=None):
         return details
 
     if has_done_signal(row):
-        logger.fdebug(
-            "[RECOVERY-CLASSIFY] %s absent in client BUT done-signal present "
-            "(history likely evicted while down) -> complete, NOT gone." % rkey
-        )
+        if raw_state == "unprobeable":
```
frankieramirez, line 658: A reconstructed NZBGet row with no NZBID and no nzblog takes this branch, gets marked downloaded, and goes onto PP_QUEUE with no folder. The #734 test for that same shape still injects unreachable and expects manual_review. If this should keep matching absent, add a replay that hits `_nzbget_history` with an empty payload so the built-in path can't drift.

azuretek:
Flagging a merge-order dependency rather than a problem with this change: #833 routes more rows into `has_done_signal()`, and that function's `issues.Status` branch cannot currently return True. #831 is the fix for it.

`classify_details` now sends `unprobeable` through the done-signal guard, so for a reconstructed row the verdict turns on `has_done_signal()`. That calls `_issue_post_processed()`, which asks:

```python
select(issues.c.Status).where(issues.c.IssueID == str(issueid))
...
return bool(rec) and rec["Status"] == "Post-Processed"
```

Nothing writes `Post-Processed` to `issues`. On `main` the only two writers both put it on `snatched` and put `Downloaded` on the library row:

- `postprocessor.py:4184-4197`, same block, two tables: `issues` gets `{"Status": "Downloaded", "Location": filename}`, then `snatched` gets `{"Status": "Post-Processed"}`.
- `updater.foundsearch:1470-1491`: `snatched` gets `downstatus` (`Post-Processed` on the PP pass), while `storyarcs`, `annuals` and `issues` each get `Downloaded`.

So `grep -rn "Post-Processed" comicarr/` returns no write to `issues`, `annuals` or `storyarcs` at all. The branch is dead, and an `unprobeable` row whose journal stage is still `snatched` falls past it to the `nzblog` fallback, or for a synthetic-HIGHCOUNT one-off past that too, since nzblog-presence is advisory for those.

Two shapes are affected once this merges:

- **One-offs.** `_library_status_done` is the only done-signal they can have while stage is `snatched`, so they stay UNKNOWN and re-probe every restart.
- **Annuals.** Real ComicVine ids routinely exceed the 900000 synthetic floor, so `is_synthetic_oneoff` reads True and suppresses the nzblog fallback. Their completion is written to `annuals`, which an issues-only lookup never reads.

#831 replaces that lookup with `_library_status_done`, which reads every table that can carry the row via `_library_rows` and accepts the statuses post-processing actually writes. It stands on its own, but it is what makes this PR's new fall-through do anything for those rows, so it wants to land first or alongside.

Happy to rebase #831 onto this branch if you would rather carry them as one.

frankieramirez:
Good catch. We'll merge #831 first, then rebase this on top so the new unprobeable fall-through actually sees a real done-signal. No need to rebase #831 onto this branch.

PR #835, opened 2026-09-02
fix: release APILOCK on every exit from the manga post-process

### LINKED ISSUE #834: One misplaced manga series leaks APILOCK and freezes all imports until restart
## Summary

A single manga series whose folder is misplaced does not just fail its own import — it **freezes every import in the app**, including the Folder Monitor that exists to rescue stuck downloads. Recovery requires a restart.

`_process_manga()` acquires `APILOCK` but releases it only on its success path. Six early returns precede that release, and each one leaks the lock permanently.

## Root cause

`Process.__init__` takes the lock when `apicall=True` — `comicarr/postprocessor.py`:

```python
if comicarr.APILOCK.locked():
    return {"status": "IN PROGRESS"}

if apicall is True:
    self.apicall = True
    comicarr.APILOCK.acquire()
```

`_process_manga()` releases it near the end, on success only:

```python
if self.apicall is True:
    try:
        comicarr.APILOCK.release()
    except Exception:
        pass
```

But every early return in that method sits *above* it:

| Line | Bail-out |
|---|---|
| 4037 | `Cannot find manga series in database` |
| 4050 | series-name guard |
| 4064 | (early guard) |
| 4071 | `No manga destination directory configured` |
| 4079 | `Series folder is outside manga destination — refusing to write` |
| 4089 | `makedirs` failure |

Each is `return self.queue.put(self.valreturn)`, taken before the release. The lock is never given back.

### Why the caller's safety net cannot save it

`comicarr/app/downloads/service.py`:

```python
outcome, pp = _run_owned_postprocess(item)
...
if pp is not None:
    try:
        if pp["mode"] == "stop" and comicarr.APILOCK.locked():
            comicarr.APILOCK.release()
    except (KeyError, TypeError):
        pass
if comicarr.APILOCK.locked():
    logger.info("Another item is post-processing still...")
    time.sleep(15)
```

`pp` comes from `return "processed", pprocess.post_process()`. For manga, `post_process()` does `return self._process_manga()`, and **every** exit in that branch is `return self.queue.put(...)` — which returns `None`. So `pp` is always `None` for manga, the release block is skipped unconditionally, and the net never fires for this branch at all.

## Observed

One misplaced series (a manga whose `ComicLocation` was still under the comics root) took down the entire pipeline for 85 minutes:

```
18:13:15 ERROR  [POST-PROCESSING] Series folder /azuredata1/media/Comics/Berserk (2003)
                is outside manga destination /azuredata1/media/Manga
18:13:15 INFO   Another item is post-processing still...
```

After that point no `Now loading post-processing command` line ever appeared again — the worker silently re-queued every subsequent item forever. And the safety net designed for exactly this bailed on the same lock, every 5 minutes:

```
18:17:19  [FOLDER-CHECK] Unable to initiate folder monitor as another
18:22:19   process is currently using it or using post-processing.
18:27:19   ...
18:32:19   ...
```

18 completed volumes sat on disk, fully downloaded at 100% health, with 0 imported.

## Reproduce

1. Take a manga-typed series whose `ComicLocation` is outside `manga_destination_dir` (any ComicVine series reclassified as manga qualifies — `switch_to_manga()` does not relocate the folder).
2. Let one of its downloads complete so post-processing runs with `apicall=True`.
3. `_process_manga()` logs `is outside manga destination` and returns at line 4079.
4. `APILOCK` is now held forever. Every later post-process item is re-queued; Folder Monitor bails every cycle. Only a restart clears it.

## Why it survived

The existing tests mock `APILOCK`:

```python
mock_apilock = MagicMock()
mock_apilock.locked.return_value = False
```

A mock's `locked()` returns whatever it was told, so a leak is invisible through it. A real lock is required to see this at all.

## Proposed fix

Guarantee the release on every exit rather than at each `return`, so future early returns are covered too. Moving the body to `_process_manga_body()` and making `_process_manga()` a thin `try/finally` wrapper does that without re-indenting the body:

```python
def _process_manga(self):
    try:
        return self._process_manga_body()
    finally:
        if self.apicall is True and comicarr.APILOCK.locked():
            try:
                comicarr.APILOCK.release()
            except Exception:
                pass
```

Releasing there is safe: the worker loop refuses to start an item while the lock is held, so if it is still held when this returns, it belongs to this call. The body's own success-path release can stay, leaving success-path timing untouched — the `finally` simply becomes a no-op there.

## Verified

| Path | Before | After |
|---|---|---|
| Series folder outside manga destination | lock **held** | released |
| Missing series row | lock **held** | released |
| No manga destination configured | lock **held** | released |
| Success path | released once | released once (no double-release) |
| `apicall=False` | never acquired | never released |

Suite on the branch: **2841 passed, 8 failed** — the same 8 failures present on pristine `main` (`tests/integration/test_schema_migration_dialects.py`), so zero new failures. `ruff check` and `ruff format --check` pass on both changed files.

## Related

The misplacement that triggers this is itself worth addressing — `switch_to_manga()` writes content-kind and search settings but never relocates the series folder, and there is no API route that can, so any ComicVine series reclassified as manga is primed for this. That is a separate issue; this one is about a local failure taking the whole app down with it.

azuretek:
Fixes #834.

## Problem

`_process_manga()` acquires `APILOCK` (via `Process.__init__` with `apicall=True`) but releases it only on its success path. Six early returns sit above that release — missing series row, no manga destination, series folder outside the manga destination, `makedirs` failure, and two guards — and each one is `return self.queue.put(self.valreturn)`, taken before the lock is given back.

The caller cannot clean up after them either. `downloads/service.py` gates its safety net on `if pp is not None`, and `pp` comes from `pprocess.post_process()` — which for manga returns `self._process_manga()`, whose every exit is `queue.put(...)`, i.e. `None`. So the net is skipped unconditionally for this entire branch.

The blast radius is the app, not the series. `APILOCK` is global and the post-processing worker takes it per item, so one leak stops every later import — and the Folder Monitor that exists to rescue stuck downloads bails on the same lock (`Unable to initiate folder monitor as another process is currently using it`). One misplaced series froze the pipeline for 85 minutes with 18 fully-downloaded volumes sitting unimported.

## Change

Moved the body to `_process_manga_body()` and made `_process_manga()` a thin wrapper that releases in a `finally`.

- **A `finally`, not a release per `return`** — that keeps the guarantee at one site and covers early returns added later. Six sites that each must remember to release is what produced the bug.
- **The body's own success-path release is left in place**, so success-path timing is unchanged; the `finally` checks `locked()` first and is a no-op there.
- **No re-indentation** — the body moves to a new method name, so the diff stays readable.

Releasing in the wrapper is safe: the worker loop refuses to start an item while the lock is held, so if it is still held when this returns, it belongs to this call.

## Test

`tests/unit/test_manga_postprocessor.py` gains `TestMangaPostProcessAlwaysReleasesApilock`, using a **real `ThreadSafeLock`**. That detail is the point — the existing `_make_pp()` helper does:

```python
mock_apilock = MagicMock()
mock_apilock.locked.return_value = False
```

A mock's `locked()` returns whatever it was told, so the leak is invisible through it. The tests assert the precondition (`__init__` really did acquire) before exercising the bail-out.

Each test also asserts **which** bail-out it exercised, via `pp.log`. The early returns are ordered — missing row → missing dir → no files → no destination → outside destination → makedirs — so a test that forgets to seed a `.cbz` silently stops at *"No manga files"* and proves nothing about the branch named in its title. (That is not hypothetical: my first draft of these tests did exactly that.)

Without the fix, four fail with the production symptom:

```
E   AssertionError: APILOCK leaked; every later import would block

FAILED ...::test_series_folder_outside_manga_destination_releases_the_lock
FAILED ...::test_no_manga_destination_configured_releases_the_lock
FAILED ...::test_missing_series_row_releases_the_lock
FAILED ...::test_no_manga_files_found_releases_the_lock
4 failed, 2 passed
```

With the fix: `6 passed`.

The two that pass either way are deliberate controls, so the change is distinguishable from just dropping a guard:

- **success path still releases exactly once** — proves the wrapper does not double-release (`RuntimeError`)
- **`apicall=False` never acquires** — proves the wrapper does not release a lock it does not own

## Verification

| | Result |
|---|---|
| Full suite, this branch | **2842 passed**, 8 failed, 1 skipped |
| Full suite, pristine `main` | 2836 passed, **8 failed**, 1 skipped |
| Delta | +6 passing (the new tests), **0 new failures** |

The 8 failures are pre-existing on `main` (`tests/integration/test_schema_migration_dialects.py`) and unrelated. `ruff check` and `ruff format --check` pass on both changed files.

### DIFF
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -4021,11 +4021,45 @@ class PostProcessor(object):
                 return self.queue.put(self.valreturn)
 
     def _process_manga(self):
+        """Post-process a downloaded manga file, always releasing APILOCK.
+
+        The real work is in _process_manga_body(); this wrapper exists solely
+        to guarantee the lock this Process acquired is released on EVERY exit.
+
+        Why a finally rather than a release at each return: the body has six
+        early returns that all precede its success-path release, and every one
+        of them leaked the lock. The caller cannot clean up after them either
+        -- downloads/service.py gates its safety net on `if pp is not None`,
+        and every exit here is `return self.queue.put(...)`, which is None.
+
+        A leak is not confined to the failing series. APILOCK is global and the
+        post-processing worker takes it per item, so one leak stops ALL
+        imports, and the Folder Monitor that exists to rescue them bails on the
+        same lock ("Unable to initiate folder monitor as another process is
+        currently using it"). A single misplaced manga series froze the whole
+        pipeline for 85 minutes.
+
+        Releasing here is safe: the worker loop refuses to start an item while
+        the lock is held, so if it is still held when this returns, it is ours.
+        """
+        try:
+            return self._process_manga_body()
+        finally:
+            if self.apicall is True and comicarr.APILOCK.locked():
+                try:
+                    comicarr.APILOCK.release()
+                except Exception:
+                    pass
+
+    def _process_manga_body(self):
         """Post-process a downloaded manga file.
 
         Routes manga downloads to the manga destination directory,
         matches files to chapters using the manga filename parser,
         and updates chapter status to Downloaded.
+
+        Call _process_manga() instead -- that wrapper owns the APILOCK
+        guarantee for every exit path below.
         """
         module = self.module
 
--- a/comicarr/postprocessor.py
+++ b/comicarr/postprocessor.py
@@ -4041,6 +4041,22 @@ class PostProcessor(object):
 
         Releasing here is safe: the worker loop refuses to start an item while
         the lock is held, so if it is still held when this returns, it is ours.
+
+        This is the ONLY release site, deliberately. The body used to release
+        itself on the success path and then carry on -- journal "moved", the
+        nzblog deletes, journal "post_processed" -- with the lock already
+        given up. Keeping that release alongside this finally would leave two
+        sites and a real window between them: PostProcessor.__init__ gates
+        only on APILOCK.locked(), and startup recovery re-drives items inline
+        on the main thread with apicall=True (app/downloads/recovery.py) while
+        the PPPOOL worker is running. An acquirer landing in that window would
+        then have its lock released by this finally, and the two would overlap.
+        With one site the lock also covers the journal write, which is where
+        it belonged in the first place.
+
+        self.apicall is part of the guard, not redundant with locked(): an
+        apicall=False Process never acquired, so a held lock there is someone
+        else's and must not be released.
         """
         try:
             return self._process_manga_body()
@@ -4255,12 +4271,6 @@ class PostProcessor(object):
         else:
             self._log("Manga post-processing complete: 0 files matched to chapters")
         logger.info("%s Manga post-processing complete for %s: %d files" % (module, series_name, processed))
-
-        if self.apicall is True:
-            try:
-                comicarr.APILOCK.release()
-            except Exception:
-                pass
 
         if processed > 0:
             self._journal_pp("moved", issueid=last_matched_issueid)
--- a/tests/unit/test_manga_postprocessor.py
+++ b/tests/unit/test_manga_postprocessor.py
@@ -697,10 +697,34 @@ class TestMangaPostProcessAlwaysReleasesApilock:
     outside destination -> makedirs), so a test that forgets to seed a .cbz
     silently stops at "No manga files" and proves nothing about the branch
     named in its title.
+
+    _run_manga counts releases rather than only reading locked() at the end.
+    ThreadSafeLock.release() swallows the RuntimeError a double release would
+    raise, so a released-twice lock and a released-once lock are both simply
+    unlocked -- the count is the only way to tell them apart, and the whole
+    point of the single-release invariant is that there must be exactly one.
     """
 
-    def _run_manga(self, tmp_path, comic_row, manga_dest, seed_file=True):
+    def _run_manga(
+        self,
+        tmp_path,
+        comic_row,
+        manga_dest,
+        seed_file=True,
+        apicall=True,
+        select_one=None,
+        expect_exc=None,
+        lock_after_init=False,
+    ):
         real_lock = comicarr.ThreadSafeLock()
+        releases = []
+        raw_release = real_lock.release
+
+        def counting_release():
+            releases.append(True)
+            return raw_release()
+
+        real_lock.release = counting_release
 
         if seed_file:
             # Required to get PAST the "No manga files found" return.
@@ -722,19 +746,30 @@ class TestMangaPostProcessAlwaysReleasesApilock:
                 nzb_folder=str(tmp_path),
                 comicid="42",
                 queue=MagicMock(spec=queue.Queue),
-                apicall=True,
+                apicall=apicall,
             )
-            # Process(apicall=True) takes the lock; that is the precondition.
-            assert real_lock.locked() is True, "expected __init__ to acquire APILOCK"
+            if apicall:
+                # Process(apicall=True) takes the lock; that is the precondition.
+                assert real_lock.locked() is True, "expected __init__ to acquire APILOCK"
+
+            if lock_after_init:
+                # Somebody else takes APILOCK after this Process was built.
+                # It has to happen here rather than before: __init__ returns a
+                # dict when it finds the lock held, which Python rejects.
+                real_lock.acquire()
 
             with (
                 patch("comicarr.postprocessor.get_manga_destination", return_value=manga_dest),
                 patch("comicarr.postprocessor.db") as mock_db,
             ):
-                mock_db.select_one.side_effect = [comic_row, None, None, None]
-                pp._process_manga()
+                mock_db.select_one.side_effect = select_one if select_one is not None else [comic_row, None, None, None]
+                if expect_exc is not None:
+                    with pytest.raises(expect_exc):
+                        pp._process_manga()
+                else:
+                    pp._process_manga()
 
-            return real_lock, pp
+            return real_lock, pp, releases
 
     def test_series_folder_outside_manga_destination_releases_the_lock(self, tmp_path):
         """The real 2026-09-01 incident: a manga series whose ComicLocation sat
@@ -744,7 +779,7 @@ class TestMangaPostProcessAlwaysReleasesApilock:
         outside = tmp_path / "Comics" / "Berserk (2003)"
         outside.mkdir(parents=True)
 
-        lock, pp = self._run_manga(
+        lock, pp, releases = self._run_manga(
             tmp_path,
             {"ComicName": "Berserk", "ComicLocation": str(outside)},
             str(manga_dest),
@@ -752,9 +787,10 @@ class TestMangaPostProcessAlwaysReleasesApilock:
 
         assert "outside manga destination" in pp.log, "did not reach the location refusal"
         assert lock.locked() is False, "APILOCK leaked; every later import would block"
+        assert releases == [True], "expected exactly one release"
 
     def test_no_manga_destination_configured_releases_the_lock(self, tmp_path):
-        lock, pp = self._run_manga(
+        lock, pp, releases = self._run_manga(
             tmp_path,
             {"ComicName": "Berserk", "ComicLocation": str(tmp_path)},
             None,
@@ -762,19 +798,21 @@ class TestMangaPostProcessAlwaysReleasesApilock:
 
         assert "No manga destination directory configured" in pp.log
         assert lock.locked() is False
+        assert releases == [True], "expected exactly one release"
 
     def test_missing_series_row_releases_the_lock(self, tmp_path):
-        lock, pp = self._run_manga(tmp_path, None, str(tmp_path / "Manga"))
+        lock, pp, releases = self._run_manga(tmp_path, None, str(tmp_path / "Manga"))
 
         assert "Cannot find manga series in database" in pp.log
         assert lock.locked() is False
+        assert releases == [True], "expected exactly one release"
 
     def test_no_manga_files_found_releases_the_lock(self, tmp_path):
         """The earliest bail-out reachable with a valid series row."""
         manga_dest = tmp_path / "Manga"
         manga_dest.mkdir()
 
-        lock, pp = self._run_manga(
+        lock, pp, releases = self._run_manga(
             tmp_path,
             {"ComicName": "Berserk", "ComicLocation": str(manga_dest / "Berserk")},
             str(manga_dest),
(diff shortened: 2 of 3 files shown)


comicarr/postprocessor.py
```diff
@@ -4021,11 +4021,45 @@ def nzb_or_oneoff_pp(self, tinfo=None, manual=None):
...
+        currently using it"). A single misplaced manga series froze the whole
+        pipeline for 85 minutes.
+
+        Releasing here is safe: the worker loop refuses to start an item while
+        the lock is held, so if it is still held when this returns, it is ours.
+        """
+        try:
+            return self._process_manga_body()
+        finally:
+            if self.apicall is True and comicarr.APILOCK.locked():
```
frankieramirez, line 4064: Keeping the body's own release at 4252 alongside this finally leaves two release sites, and the gap between them is real. The body releases, runs the journal "moved" transition and the nzblog deletes, then lands here. If another apicall=True post-process grabs the lock in that gap, this releases it out from under them and the two overlap. That second acquirer exists: startup recovery re-drives items inline on the main thread via `process.Process(..., apicall=True)` (recovery.py:580-602) while the PPPOOL worker is already running and its only gate is `locked()`. Before this PR there was one release and no re-check, so the window is new. I'd delete 4252-4256 and let this finally be the only release; the lock then also covers the journal write, which is where it belongs anyway.

comicarr/postprocessor.py
```diff
@@ -4021,11 +4021,45 @@ def nzb_or_oneoff_pp(self, tinfo=None, manual=None):
...
+        imports, and the Folder Monitor that exists to rescue them bails on the
+        same lock ("Unable to initiate folder monitor as another process is
+        currently using it"). A single misplaced manga series froze the whole
+        pipeline for 85 minutes.
+
+        Releasing here is safe: the worker loop refuses to start an item while
+        the lock is held, so if it is still held when this returns, it is ours.
+        """
+        try:
+            return self._process_manga_body()
```
frankieramirez, line 4062: The raise path is the one exit a release-per-return could never cover, and it's also the one exit with no test. Something like `mock_db.select_one.side_effect = RuntimeError("db down")`, `with pytest.raises(RuntimeError): pp._process_manga()`, then assert the lock is free, would pin it.

tests/unit/test_manga_postprocessor.py
```diff
@@ -601,3 +601,116 @@ def test_unmatched_file_logs_warning(self, tmp_path):
...
+        series.mkdir(parents=True)
+        (tmp_path / "Berserk v16.cbz").write_bytes(b"fake cbz")
+
+        lock = self._run_manga(
+            tmp_path,
+            {"ComicName": "Berserk", "ComicLocation": str(series)},
+            str(manga_dest),
+        )
+
+        assert lock.locked() is False
```
frankieramirez, line 871: This can't tell one release from two, and the RuntimeError reasoning in the docstring doesn't hold: `ThreadSafeLock.release()` swallows RuntimeError. I deleted the body's release at 4252 and separately made the wrapper release unconditionally, and all five tests stayed green both times. Wrapping `real_lock.release` with a counter in `_run_manga` and asserting it fired exactly once would make this a real control.

tests/unit/test_manga_postprocessor.py
```diff
@@ -601,3 +601,116 @@ def test_unmatched_file_logs_warning(self, tmp_path):
...
+        real_lock = comicarr.ThreadSafeLock()
+        with patch.object(comicarr, "APILOCK", real_lock), patch.object(comicarr, "CONFIG", MagicMock()):
+            PostProcessor(
+                nzb_name="x.cbz",
+                nzb_folder=str(tmp_path),
+                comicid="42",
+                queue=MagicMock(spec=queue.Queue),
+                apicall=False,
+            )
+        assert real_lock.locked() is False
```
frankieramirez, line 993: This only exercises `__init__`. `_process_manga()` is never called, so the `self.apicall is True` half of the finally guard has no coverage; dropping it to `if comicarr.APILOCK.locked():` also leaves all five tests passing. A test that constructs with apicall=False, then takes the real lock, then calls `pp._process_manga()` with `comic_row=None` and asserts the lock is still held would cover it.

tests/unit/test_manga_postprocessor.py
```diff
@@ -601,3 +601,116 @@ def test_unmatched_file_logs_warning(self, tmp_path):
...
+            with (
+                patch("comicarr.postprocessor.get_manga_destination", return_value=manga_dest),
+                patch("comicarr.postprocessor.db") as mock_db,
+            ):
+                mock_db.select_one.side_effect = [comic_row, None, None, None]
+                pp._process_manga()
+
+            return real_lock
+
+    def test_series_folder_outside_manga_destination_releases_the_lock(self, tmp_path):
```
frankieramirez, line 774: `_run_manga` never writes a .cbz under tmp_path, so this test and `test_no_manga_destination_configured_releases_the_lock` bail at the `if not manga_files` guard before the containment check or `get_manga_destination()` ever run. The lock assertion still passes since that's also an early return, but the incident branch this is named after doesn't execute (I printed `pp.log`; both end with "No manga files found"). Writing the fake cbz in `_run_manga` like the success test does, and asserting on the last log line, would fix it.

comicarr/postprocessor.py
azuretek: Agreed on all of it, and done your way in eb2ad89 — deleted 4252-4256, the `finally` is now the only release site, and the lock covers the journal write.

I checked the acquirer you named before changing anything. `__init__` gates on `APILOCK.locked()` alone (postprocessor.py:130) and startup recovery constructs `process.Process(..., item["apicall"], ...)` inline on the main thread inside its lease, so yes — an acquirer can land between the body's release and the wrapper's, and the wrapper would then release a lock it no longer owns. And you're right that the window is new here: with one release and no re-check there was nothing to steal.

What this exposed is that my `test_success_path_still_releases_exactly_once` could not have caught it, and neither could a release counter, which is what I reached for first. Once the body releases, the wrapper's `locked()` check just skips its own release — so the count is 1 either way. Restoring 4252-4256 with a counting `release` in place still passed.

The property that actually distinguishes them is the ordering, so `test_lock_is_still_held_during_the_journal_write` wraps `pp._journal_pp` and records `APILOCK.locked()` at each transition:

```python
assert all(held for _, held in held_during), (
    "APILOCK was already released during %s" % [stage for stage, held in held_during if not held]
)
```

That one does fail with 4252-4256 restored. Ran it as a mutation matrix — restore the body release / release unconditionally / drop the `apicall` half of the guard / remove the `finally` entirely — and each now fails at least one test. Full suite is 2830 passing.

comicarr/postprocessor.py
azuretek: Added in eb2ad89 as `test_raising_body_still_releases_the_lock`, close to what you sketched:

```python
lock, pp, releases = self._run_manga(
    tmp_path, None, str(tmp_path / "Manga"),
    select_one=RuntimeError("db down"),
    expect_exc=RuntimeError,
)
assert lock.locked() is False, "APILOCK leaked out of a raising post-process"
assert releases == [True], "expected exactly one release"
```

`_run_manga` grew a `select_one` override and an `expect_exc` so the `pytest.raises` can sit inside the patch context and the lock is still inspectable afterwards.

Worth noting it is a worse exit than the six returns, not just an uncovered one: it leaves no log line, so there is nothing in the log to notice the leak by. Removing the `finally` fails this test.

tests/unit/test_manga_postprocessor.py
azuretek: You're right, and thanks for actually running the mutations — the docstring's reasoning was wrong. `ThreadSafeLock.release()` catches `RuntimeError` and passes (`comicarr/__init__.py:96-104`), so "the wrapper must not double-release (RuntimeError)" describes an exception that cannot be raised, and `locked() is False` is true after one release or two.

I added the counter you suggested, and it caught the two guard mutations (unconditional release, and dropping the `apicall` half) — but not the one you were pointing at. With the body's release still in place the count is *still* 1, because the wrapper's `locked()` check skips its own release once the body has released. So the counter alone would have let 4252-4256 stand.

Kept it — it pins "exactly once" cheaply across all six bail-outs — but the double-release is caught by the ordering assertion in `test_lock_is_still_held_during_the_journal_write` instead, per the other thread. Renamed this one to `test_success_path_releases_exactly_once` and rewrote the docstring, since the old one asserted a RuntimeError that never happens.

tests/unit/test_manga_postprocessor.py
azuretek: Correct — it never called `_process_manga()`, so the `finally` never ran and the `self.apicall is True` half was uncovered. Rewritten in eb2ad89 as `test_apicall_false_does_not_release_a_lock_it_does_not_own`, following your sketch: construct with `apicall=False`, take the real lock, call `_process_manga()` with `comic_row=None`, assert the lock is still held.

One wrinkle worth flagging — the lock has to be taken *after* construction, not before. `PostProcessor.__init__` does `return {"status": "IN PROGRESS"}` when it finds APILOCK held (postprocessor.py:130-131), and Python rejects a non-None return from `__init__` with `TypeError: __init__() should return None, not 'dict'`. So constructing while the lock is held raises rather than returning the sentinel. Same ordering either way for this test, but that IN PROGRESS guard looks like it cannot work as written — happy to file it separately if you want it looked at; it's outside this branch.

Both mutations you named now fail: dropping the guard to `if comicarr.APILOCK.locked():`, and making the wrapper release unconditionally.

tests/unit/test_manga_postprocessor.py
azuretek: This one was already fixed before your review landed — the comment is anchored to c6a0e9a, and the branch head at the time you commented (6387054) had the seeding. `_run_manga` takes `seed_file=True` and writes `Berserk v16.cbz` under `tmp_path`, with `test_no_manga_files_found_releases_the_lock` passing `seed_file=False` to reach that guard deliberately.

Your diagnosis was exactly right though, and it is why the class docstring now spells out the ordering of the bail-outs and every test asserts which one it hit:

```python
assert "outside manga destination" in pp.log, "did not reach the location refusal"
```

Same fix, arrived at the same way — by printing `pp.log` and finding "No manga files found" where the location refusal should have been.

azuretek:
Rebased onto current main and force-pushed (previous head kept at `azuretek/comicarr@backup/apilock-20260902`).

CI went red on the rebase, and the cause is worth naming because it is not the lock change: main has since gained `persist_manga_location_if_needed`, which repoints a manga series still pointing at the comics tree *before* post-processing reaches the destination check. That was the exact input `test_series_folder_outside_manga_destination_releases_the_lock` used, so the branch it was named after is no longer reachable for it — correctly, since that case is now repaired rather than refused.

The refusal still stands where the repair cannot compute a replacement, so the test drives that instead (a series with no usable name leaves the outside path in place). Added a companion for the repaired path, which is now the common case and must not leak the lock either.

Worth flagging for its own sake: that repair narrows the refusal to inputs the repair declines, so anything else asserting on it is probably now testing a dead branch.

